PR #1, opened 2015-05-03
Trigger Host Alert

lorenzobrunetti:
Trigger Host Details is now under the alert section and appear only for
remote hosts


lucaderi:
Some comments
1. The code you sent me isn't in sync with the repository: please update your code first.
2. The alerts should be generated only by local hosts (not the other way round).
3. When you define NetworkInterface::getdumpFlowPreference() please capitalise names (i.e. NetworkInterface::getDumpFlowPreference) and make it mode meaningful. As this method returns a boolean your should not use getXXX as it expects a value back
4. For methods such as those you defined, please make them inline.
Please fix the above issues and try again submitting your contribution

PR #37, opened 2015-05-28
Fixed flow graphical categorization

mloporchio:
Fixed an issue which caused flows with ID = “::1” and host_server_name = “localhost” to be displayed in alerts page and classified as “malware”.


lucaderi:
@mloporchio This merge request is no longer necessary as a fix in the Host class has been already added that should take care of the issue you planned to tackle. Thanks for your contribution anyway.

PR #41, opened 2015-06-02
Changes needed for ntopng to compile on OpenBSD

zhouska:
Hi Luca, 

commit 71414f65be84634ce04881a909bca235d6f4a0df is needed for ntopng to compile on OpenBSD, remaining changes address lack of '-ldl' in OpenBSD and various small changes relating to security (these are specific to OpenBSD).

Zbynek


lucaderi:
@zhouska Thanks for the contribution. Glad to read that ntopng now works on OpenBSD too

PR #64, opened 2015-06-26
Re-enable clang compiler.

GDvalle:
Current FreeBSD 10.1 is failing to compile with GCC, and reverting this
commit is the only thing necessary for a successful build.

Ubuntu 14.04 with GCC is producing a successful build, but I've had the
binary segfault just after starting.  After reverting this, and
installing clang, builds are successful and do not segfault.

This reverts commit 37f407a7ccd9e0c4df17fb7ca9664781e4c14dce.

### DIFF
--- a/configure.seed
+++ b/configure.seed
@@ -287,12 +287,11 @@ then
 	GMAKE="make"
 fi
 
-GPP=`which g++`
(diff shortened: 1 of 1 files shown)


lucaderi:
Thanks

PR #71, opened 2015-07-05
Add support for libpcap Raw IP Link-Layer header type

VittGam:
Hello,

This patch allows ntopng to work with Linux TUN/TAP devices in TUN mode.

Best regards,
Vittorio G


lucaderi:
Thank you

VittGam:
You're welcome! But you might want to check #72 as the 101 is actually wrong here...

PR #76, opened 2015-07-07
src/HistoricalInterface.cpp: create flowsManager instance for the Historical interface

VittGam:
Hello,

This commit avoids a segmentation fault when opening the Flows page for the Historical interface in the web interface.

It happened because the `flowsManager` was not being initialized for the `HistoricalInterface`.

So when the `FlowsManager::select` function was trying to call `intf->get_flows_hash()->walk(flows_select_walker, (void*)&info);`, `intf` could not be retrieved because `this` was `null`, and the program segfaulted.

Best,
Vittorio G


ariava:
@VittGam good catch, thanks!

VittGam:
You're welcome @ariava!

PR #79, opened 2015-07-09
src/PacketDumper.cpp: correctly set interface datalink type for the PacketDumper.

VittGam:
Hello,

This commit fixes packet dumping for pcap interfaces such as TUN/TAP in TUN mode, ipip tunnels, sit (IPv6 over IP) tunnels.

Best regards,
Vittorio G


ariava:
@VittGam Thank you!

PR #83, opened 2015-07-14
fixed init script exit code

Holmser:
Init script is exiting with 0 when ntopng is not running.  Fixed to exit with 1,


lucaderi:
Thank you

PR #95, opened 2015-07-26
ntopng batman-adv dissector

beratx:
This patch aims to add support to ntopng to visualize antennas of mesh networks which use batman-adv routing protocol version 2013 or 2014.


lucaderi:
Thank you

PR #127, opened 2015-08-14
added function to get max speed of eth - modified interface speed page

kYroL01:
added: 
ntop_get_speed_max in Lua.cpp
getMacSpeed in Utils.h and Utils.cpp
getSpeedMax in lua_utils.lua


lucaderi:
Thank you

PR #128, opened 2015-08-18
fixed warning in zmq according to update of libzmq

kYroL01:
issue caused by use of tempnam fixed in https://github.com/kYroL01/libzmq/blob/master/src/ipc_listener.cpp#L149

also read this: http://lists.zeromq.org/pipermail/zeromq-dev/2013-December/024223.html


lucaderi:
This fix is not necessary for two reasons: the latest zeromq (now included in ntopng) includes it and also because accepting it is a bad idea as the fix should go into zeromq and not in ntopng. For this reason we're unable to accept your contribution.

PR #131, opened 2015-08-20
fixing missed host part in url http packet request

kYroL01:


lucaderi:
This work is no longer necessary as the fix for the problem is already in git.

PR #140, opened 2015-08-30
build: fix MAN_DIR

dunn:
Was broken for Darwin in cefc119de1dee02a01b7df983d63ca0b0588cda7.

Also sets MAN_DIR to the mandir itself rather than its parent.


lucaderi:
Thanak you

Issue #148, opened 2015-09-02
Issue with "-i view:" parameter


gianogli:
I think that ntopng have a problem to manage the "-i view:" parameter (used to aggregate more interfaces in a single view).

If I use the "view" option in this way:
    ntopng -i eth6 -i eth7 -i view:eth6,eth7 [...]
all works fine. If I use instead the zc drivers:
    ntopng -i zc:2@0 -i zc:2@1 -i view:zc:2@0,zc:2@1 [...]
the virtual aggregated interface "zc:2@0,zc:2@1" doesn't appear.

I think that the problem is related to the "NetworkInterfaceView::NetworkInterfaceView(const char *name, int id)" function and should be caused by ":" character used both by the "view" option and by the name of the zc interfaces.

Thanks...

   Gianluca

kYroL01:
Please @gianogli test if the issue is fixed.

lucaderi:
Just fixed

PR #151, opened 2015-09-05
fixed wrong '-i view' for zc option

kYroL01:
As evidenced in https://github.com/ntop/ntopng/issues/148 , the aggregation of more interfaces in a single view does not show caused missing parsing of 'zc' option.


lucaderi:
This fix is not general but it solves only the zc: problem with view: and thus it cannot be accepted as is.

Issue #153, opened 2015-09-07
How to reset password on ntopng?


Drewshg312:
Hi!
I was changing my password and forgot it)
How do I reset my pass? The username is admin.

    redis-cli del user.admin.password

Did not help. Neither reinstallation helps (using homebrew on os x 10.10.5) ((

lmangani:
Hi,

Please try using ```redis-cli SET user.admin.password <MD5>```

If you want to drop all data and use redis for ntopng exclusively use ```redis-cli flush all ```

Drewshg312:
I've tried `redis-cli SET user.admin.password MD5_of_newpass`  too. No results.
and 

```
$ redis-cli flush all
(error) ERR unknown command 'flush'

what am I doing wrong?

lucaderi:
https://www.ntop.org/guides/ntopng/faq.html#cannot-login-into-the-gui

Drewshg312:
I've already read it and done everything like it says with no luck… That's why I'm asking you here

Drewshg312:
There you go! 
Found the answer here **(in comments)**:

http://blog.redbranch.net/2013/12/12/reset-ntopng-admin-password/

"I tried the above commands without any luck. I determined the user account above was not correct by running strings /var/lib/redis/dump.rdb | grep admin which showed a key for ntopng.user.admin.password. I was able to reset the password then by doing

redis-cli SET ntopng.user.admin.password ea847988ba59727dbf4e34ee75726dc3

**It should also be noted that ntopng does not allow special characters in the passwords and silently fails. This is how I got locked out of the UI**"

This should be added to https://github.com/ntop/ntopng/blob/dev/doc/README.ntopng … if anyone cares…

lucaderi:
Please send us a merge request for the read me and open an issue for the issue you identified

lmangani:
Docs updated by ntop team

Drewshg312:
would be great to have some hint in an actual application - would save a lot of time…

lmangani:
I honestly believe the README does suffice for this sort of information as the go-to document. 
Feel free to submit a pull request with any proposal of course.

Drewshg312:
check the behavior of the application after changing the password to something like "34$#2"… Looks really buggy… just saying…

lmangani:
Our instructions do indicate how to RESET the password to its defaults, and not how to change it to a bad string. As said, improvements and contributions are super welcome any time!

Drewshg312:
[...]

PR #156, opened 2015-09-09
added check alphanumeric string for password in manage user form

kYroL01:
related to https://github.com/ntop/ntopng/issues/153


kYroL01:
Ok Luca, so there is another fix to do, because now, if a user want insert a password with a special char, the function purifyHTTP param in Lua.cpp(4616) discard any special char.
I'll try fo fix this function or do another one just for the password form.

lucaderi:
#153 already fixed. Thanks

PR #180, opened 2015-09-23
2.0 stable

paoloB132:
Checks of allowed password characters has too be the same both on login script and change password ones. Actually they are not. On the login script they are restricted using reg exp to \w, while on change password script they are not checked.
I suggest anyway to allow some other symbols ` ~ ! @ # $ % ^ & * ( ) _ - + = { } [ ] \ | : ; " ' < > , . ? / to make password stronger.


lucaderi:
Sorry I can't accept this fix: it is too complicated.

I accept fixes for the development version. Please try again

PR #183, opened 2015-09-26
fix #174: removed max cap in number input field for maximum interface speed

### LINKED ISSUE #174: Preferences: Interface Speed issue.
If I use, as input interface, a ZC "virtual" interface or an aggregated "view" interface I can't choose an "Interface Speed" value bigger than 1000 Mbit.

In this example the view interface "view:zc:2@0,zc:2@1,zc:2@2,zc:2@3"  represents a tap interface of a 1Gbit/s full duplex link. So the Interface Speed should be 2000Mbit.
ntopng_interface_speed

I've the same problem with a ZC interface obtained as a "fan-out" (zbalance_ipc command with the option -m 2) of more than two 1Gbit/s interfaces. Also in this case I can't choose an "Interface Speed" value bigger than 1000 Mbit.

Thanks.

simonemainardi:
Interface speed cannot always be guessed so it is not really the case to use a cap

### DIFF
--- a/scripts/lua/set_if_prefs.lua
+++ b/scripts/lua/set_if_prefs.lua
@@ -26,20 +26,13 @@ else
    custom_name = ntop.getCache(key)
 end
 
-key = 'ntopng.prefs.'..ifname..'.speed'
+-- ifstats.name = tostring(ifstats.name)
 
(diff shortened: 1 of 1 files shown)


lucaderi:
Thank you

PR #184, opened 2015-09-26
Flip Lat/Long for geo_map fields in Kibana

lmangani:
Corrects the "geo_point" format export for Elasticsearch/Kibana visualization.
Does not seem to impact or interfere with internal usage or hosts_geomap.lua but please doublecheck!


lucaderi:
Thank you

PR #195, opened 2015-10-04
ignore build-time generated files for third-party applications

simonemainardi:
Luca, please consider merging this. It cleans the output of git, especially when doing `git status`


lucaderi:
Thank you

PR #226, opened 2015-10-12
fixed some alerts bugs caused by no synchronization between alerts in…

kYroL01:
…fo show on tables and number of alerts in show_alerts.lua page.

Please, check if the issue is resolved. 
Thank you.


lucaderi:
As discussed, please review your patch once again and resubmit. Thanks.

PR #249, opened 2015-10-22
Reworked PF_RING configure checks

simonemainardi:
@lucaderi , I and @cardigliano have tried to rework `configure.seed` that was preventing ntopng from being compiled on CentOS 6.7

We have:
- removed NUMA from PF_RING checks -- PF_RING no longer uses it
- reworked PF_RING check branches to fix some errors,

please check this new version and consider merging it.


lucaderi:
This merge will break many things including the ARM port. It needs to be polished first.

PR #326, opened 2015-12-18
Respect HTTPPrefix for search box submit

h8h:
Right now by fireing up the search it redirects to the wrong page if ntopng is behind a reverse proxy with another http prefix.


lucaderi:
Thank you

PR #332, opened 2015-12-28
simulator interface

simonemainardi:
Hi Luca, please consider including this simulator interface. The aim of this interface is to simulate flows and is created with the -i modifier as follows:

-i simulator:XX

where XX is the number of new flows that should be added every second.


lucaderi:
Please rework your patch as the dummy interface has already been added to the repository

PR #459, opened 2016-03-24
A total hack to enable graphs in the 'history' part of interfaces to …

markfoodyburton:
This is a terrible hack, I didn't even properly test it with the latest dev branch, and I urge you not to merge it. I dont even think this belongs in this file, it should be in the graph utils, but I didn't want to touch that because I have no idea where or how thats used... So this was a quick hack. None the less, the 'hack' could potentially give some ideas about how to do this 'properly' in the graph utils file....

…track multiple interfaces. Your millage will vary, and this is a total hack, the change shouldn't even be in this file !.

### DIFF
--- a/scripts/lua/if_stats.lua
+++ b/scripts/lua/if_stats.lua
@@ -13,6 +13,612 @@ require "alert_utils"
 require "db_utils"
 
 sendHTTPHeader('text/html; charset=iso-8859-1')
+function mydrawRRD(ifid, host, rrdFile, zoomLevel, baseurl, show_timeseries,
+                 selectedEpoch, selected_epoch_sanitized, topArray)
+   local debug_rrd = false
+
+   if(ntop.isPro()) then
+      drawProGraph(ifid, host, rrdFile, zoomLevel, baseurl, show_timeseries, selectedEpoch, selected_epoch_sanitized, topArray)
+      return
+   end
+
+   ifs = {}
+   ifids = {}
+   for iname in string.gmatch(ifname, "[^,]+") do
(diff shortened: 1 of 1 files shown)


lucaderi:
@simonemainardi Please see this merge request.

simonemainardi:
@markfoodyburton thanks for the contribution.

@lucaderi, the patch attempts to chart separate lines for each interface of a view. This is already available in the pro version but not in the community. I would like to include the patch. However, the code contributed would require a refactoring. Indeed, the contributor placed in `if_stats.lua` what it is usually placed in `graph_utils.lua`. For this reason I don't feel the code can be merged in this way.

Diff of the relevant parts of the merge is enclosed.

[merge_diff.txt](https://github.com/ntop/ntopng/files/210349/merge_diff.txt)

lucaderi:
@simonemainardi If a refactoring is necessary @markfoodyburton could do that to enable un merging the patch.

markfoodyburton:
I agree - it should be done elsewhere, but actually the code in graph_utils.lua should probably be enhanced, rather than this crude thing added. Unfortunately I dont have time to work on this, and given that you already have well written code available, I dont really feel competent to edit your code, and given you have a 'clean' solution, I'm not too motivated to build a different versions. (Of course, thats your choice, but it would seem to add to the support burden for no good reason).
For now, I would see this as a patch that people can use to get the functionality from the pro version if they want, until such a time as you move the code from the pro version into the community version....

If I do get time to fiddle, I might have a play at making the patch better, but for now ......

Cheers

Mark.

PR #471, opened 2016-04-02
Fixes issues when sorting flows

### LINKED ISSUE #464: Wrong flows sort
In ntopng if you go to /lua/flows_stats.lua and click on the Info column, the sort is not global but per page. If you click on page 2, you will see that page 2 is sorted, but not a global sort just per page sort.  I believe the bug in sin flow_search_walker() inside NetworkInterface.cpp

simonemainardi:
Dear Luca, please consider including this fix in the dev branch.

I tested with success for findFlowByKey and findFlows.

I was not able to test for: findUserFlows, findPidFlows, findFatherPidFlows. I am not sure these function are still active in the code.

TODO: check also host sort that may be affected as well

Fixes #464


lucaderi:
@simonemainardi Thank you

PR #489, opened 2016-04-13
Added bool Flow::isSuspiciousFlow() function

ghiltanas:


lucaderi:
@ghiltanas The code is suboptimal as the indentation is different from the standard, you have hard-coded constants that should go to https://github.com/ntop/ntopng/blob/dev/include/ntop_defines.h, and in your changes you are modifiyng lines for no reason. Next time please pay attention to this (I am going to fix them myself)

PR #533, opened 2016-04-24
Drop preivileges after creating pid file.

madpilot78:
This allows it to also create the pid file in system directories not accessible to the user it's going to switch to.


lucaderi:
Thank you

PR #535, opened 2016-04-24
Files should go into DESTDIR if defined.

madpilot78:
As subject states, this looks like a forgotten statement, unless I'm missing something.


lucaderi:
Thank you

PR #616, opened 2016-06-16
Added CONST_MAX_HTTP_REQUEST_DELTA_TIME

ghiltanas:
New #define to check max time to complete an http request (3 sec)


lucaderi:
@ghiltanas I cannot merge this request as it includes commits that are already into ntopng. Please resync _first_ with the corrent ntopng tree then send us the merge request only for the _original_ code you have written and not for the whole code. Thank you

Issue #634, opened 2016-07-01
init.d script bug


mathieufrh:
Hi,

I think there is a bug in the ntopng init.d script (/packages/etc/init.d/ntopng).
Whenever I start it I have:

```
$ sudo service ntopng start
$ sudo service ntopng status -l
● ntopng.service - LSB: Start/stop ntopng web
   Loaded: loaded (/etc/init.d/ntopng)
   Active: active (running) since Fri 2016-07-01 09:20:43 CEST; 4s ago
  Process: 13285 ExecStart=/etc/init.d/ntopng start (code=exited, status=0/SUCCESS)
   CGroup: /system.slice/ntopng.service
           └─13294 /usr/bin/ntopng /etc/ntopng/ntopng.conf

Jul 01 09:20:41 mercure ntopng[13294]: [NetworkInterface.cpp:1057] WARNING: If you have TSO/GRO enabled, please disable it
Jul 01 09:20:41 mercure ntopng[13294]: [NetworkInterface.cpp:1059] WARNING: Use: sudo ethtool -K eth0 gro off gso off tso off
Jul 01 09:20:43 mercure ntopng[13285]: Starting ntopng:Unable to start ntopng ... failed!
Jul 01 09:20:43 mercure ntopng[13285]: failed!
```

It is active (and working) but in the script output it says "failed!".
The fact is that the get_ntopng_pid function cannot find the PID because it get the PID file from the configuration but fail to assert that it exists:

```
$ PID_FILE=$(cat /etc/ntopng/ntopng.conf | grep -v '^#' | grep -E '\-G=|--pid'|cut -d '=' -f 2)
$ echo $PID_FILE
/var/tmp/ntopng.pid
# Fails with the quotes
$ if [ -f "$PID_FILE" ]; then echo "OK"; else echo "FAIL"; fi
FAIL
# Works without the quotes
$ if [ -f $PID_FILE ]; then echo "OK"; else echo "FAIL"; fi
OK
```

lucaderi:
@mathieufrh Can you please send us a poll request with the fix to this problem?

PR #636, opened 2016-07-01
Fix init script bug (say fail but run)

mathieufrh:
See iusse #634.


lucaderi:
I added quotes to avoid empty files, put probably I have added a side effect. Let's include your patch hoping that we have addresses the open issue. Thank you

PR #660, opened 2016-07-13
Log preferences

mathieufrh:
I choose to put some files that I added in the .gitignore, I know there are not on the original project.
Basically I added a TRACE log level which I use to print hints when developing (not debugging).
I also let redis handle the log level. By default the log level is set in Trace constructor with:

```
traceLevel = TRACE_LEVEL_NORMAL
```

But if we can GET a log level from redis then this one is set in `Ntop::start`.
I also add a tab in the preferences page to let the user choose the log level. When the user click on save it simply set the choosen log level in redis and call the API function `ntop_set_logging_level` which set `Trace.traceLevel` (maybe redis cache should be set in this function too, don't konw which is the best way).


simonemainardi:
thanks for submitting @mathieufrh .

I have the following observations:
- custom log level set via preferences is not persistent across restarts: even if it is saved into redis, it does not get read upon ntopng startup
- upon startup a possible -vvvv option from the command line will take the precendence over any custom log level previously set
- I would use the following policy: 1. if -v has been specified, then use it regardless of the custom value set via ntopng preferences; 2. if -v has not been specified, then try and read the value stored in redis and possibly apply the default trace level. You can look at the auto_logout that behaves in the same way: https://github.com/ntop/ntopng/blob/dev/src/HTTPserver.cpp#L208
- I noticed you submitted a gitignore file with some logo-related excludes. Please, tell me you are not replacing the ntopng logo  ._.

mathieufrh:
Ok thank you for the answer. I will make the modifications.
Concerning the logo I just make it smaller to save some space.

simonemainardi:
thank you for the pull request. We have accepted it in the main dev branch. Please note that I have made some changes in order to make it more consistent with the current ntopng implementation. Anyway, thank you.

PR #689, opened 2016-08-12
Implement SSL/TLS handshake detection

emanuele-f:
Makes it possible to know if packet data is SSL/TLS handshake data or real payload.
Requires ntop/nDPI#244 to work with IMAP STARTTLS.


lucaderi:
Thank you

PR #684, opened 2016-08-10
Add IMAPS command detection

emanuele-f:
Enable src/Flow.cpp:1876 and script/callbacks/housekeeping.lua:6 to test

PR #690, opened 2016-08-14
Implements user activity detection framework

emanuele-f:
@lucaderi I have based this framework on these assumptions:
- each activity has 3 counters: upbytes, downbytes and background bytes
- each flow can be marked with an activity ID, which may change
- each flow can be assigned a filter function who is responsible for discriminating between real activity and background activity

I've changed "housekeeping" script name to "useractivity". I'm still using string protocol name to avoid passing to lua protocol numeric ids.

I'm aggregating activity per local host and passing activity values to a triple rrd counter, which is stored into [hostdirectory]/activity/[activityname].rrd. I still need to implement callbacks to retrieve rrd counter values to provide visualization.

I've integrated #684 changes as a filter for showcase. Looking forward for your feedback.


lucaderi:
Thank you very much for this contribution. It looks good to me and worth to include. I have seen minor things I would like to fix, but I first merge it and then do that on a different commit.

PR #694, opened 2016-08-16
Activity detection framework

emanuele-f:


httpdocs/js/cubism.v1.js
```diff
@@ -1,11 +1,7 @@
 (function(exports){
```
emanuele-f, line 1: Customized file moved to cubism_ntop.v1.js

scripts/lua/host_details.lua
```diff
@@ -711,6 +712,96 @@ end
 	    </td></tr>
       ]]
 
+-- Host activity stats
+if false then
```
emanuele-f, line 716: Put true to enable stats visualization

lucaderi:
Thank you. I will validate the code later and report if necessary.

PR #707, opened 2016-08-22
New user activities and filters

emanuele-f:


lucaderi:
Thank you

lucaderi:
I have made this fix https://github.com/ntop/ntopng/commit/ac2db07d4c0afb2e586f1fe3f20d12c18d8d245c
to avoid a crash. Please check the code logic.

lucaderi:
However there is still a bug you need to fix

23/Aug/2016 09:50:40 [NetworkInterface.cpp:3995] WARNING: Error while executing flowProtocolDetected [rc=2][...s/deri/network/ntopng/scripts/callbacks/flowactivity.lua:277: bad argument 1 to 'unpack' (table expected, got nil)]

emanuele-f:
ac2db07 it's ok, for the other bug there is a typo error in line 228 "detaults". Sorry for these.

PR #718, opened 2016-08-27
Activity detection changes

emanuele-f:
Changelog:
- Remove "App" activity
- Add separate "Facebook" activity
- Move SSL to "Other" activity by default, but allow "Web" filter to move it to "Web" activity
- Remove Yahoo protocol from "Chat" activity
- Fix default activity parameters (also causing all SMPTS traffic to be registered as background data)
- Add a "ratio" filter to exclude Facebook integration in websites from real "Facebook" activity
- Adjust some protocol activity detection parameters
- Remove "user_activity_none" from UserActivityID enum to be less error prone

### DIFF
--- a/include/ActivityFilters.h
+++ b/include/ActivityFilters.h
@@ -34,12 +34,17 @@ typedef unsigned int uint;
 class Flow;
 
 typedef enum {
-  activity_filter_all,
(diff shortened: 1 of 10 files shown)


lucaderi:
Thank you

PR #735, opened 2016-09-10
Make all flowactivity callbacks run in a single thread

### LINKED ISSUE #712: ntopng segfault with commit 11fcf54 .
Good afternoon.
Compilation of commit 11fcf54 is correct, but i got a segfault when I start ntopng :

ntopng[28048]: segfault at 4b ip 00000000004a697b sp 00007ffdb85c3540 error 4 in ntopng[400000+1c9000]

PF_RING and nDPI were updated to last commit.

Linux : 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt25-2+deb8u3 (2016-07-02) x86_64 GNU/Linux
gcc (Debian 4.9.2-10) 4.9.2

Best regards

emanuele-f:
This should solve any concurrency issues with flowactivity callbacks and fix #712. A single Lua VM is used.

### DIFF
--- a/include/NetworkInterface.h
+++ b/include/NetworkInterface.h
@@ -75,7 +75,7 @@ class NetworkInterface {
 
   /* Lua */
   bool reloadLuaInterpreter;
-  lua_State *L_flow_create, *L_flow_delete, *L_flow_update;
+  lua_State *L_flow_activity;
 
   /* Second update */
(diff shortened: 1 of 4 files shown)


lucaderi:
As previously discussed, I will take into account the joint comments in a future commit

PR #738, opened 2016-09-12
Make activity detection work with recent changes

emanuele-f:
Commit 5a27c35 is required to make things work as before.
Commit d1912ba only contains explanation notes, please apply at your will.

### DIFF
--- a/src/Flow.cpp
+++ b/src/Flow.cpp
@@ -379,8 +379,6 @@ void Flow::processDetectedProtocol() {
 						       &categorization.category,
 						       true);
 	      }
-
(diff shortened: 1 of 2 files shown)


src/Flow.cpp
```diff
@@ -379,8 +379,6 @@ void Flow::processDetectedProtocol() {
 						       &categorization.category,
 						       true);
 	      }
-
-	      iface->luaEvalFlow(this, callback_flow_proto_callback);
```
emanuele-f, line 383: This callback here is not usefull

lucaderi:
Thanks

Issue #740, opened 2016-09-13
datarate on disk


Niko78:
I use nprobe/ntop as netflow (v9) collector
I have around 45 routers (60k Hosts, 150k
flows) which send their netfow stream on this server
Bandwidth usage (input) on ntop server is less than 200 Kbps... but
ntop writes at more than 20MBps still on disk.

I have a second server with 400 routers with less hosts/flows... bandwidth usage is 500 kbps but datarate on disk is more than 50MBps !

we planned to migrate to SSD disks soon but I don't understand how is
it possible for ntop to receive at 200/500 Kbps and write at 20MBps (or even more than 50MBps) on disk ?

thanks

nico

simonemainardi:
Please enclose proof that is actually ntopng that is writing at 20/50Mbps on disk.

However, among the _27_ issues you have already filed so far (https://github.com/ntop/ntopng/issues?utf8=%E2%9C%93&q=author%3ANiko78) there is this one (https://github.com/ntop/ntopng/issues/628#issuecomment-229033070) that shows how you define one huge /16 local network. My feeling is that you have 60k local hosts (maybe also with L7 statistics enabled) and this is causing intensive I/O to disk.

Niko78:
L7 was disabled
I let you see my screenshot (less hosts/flows but high IO)

![ntop](https://cloud.githubusercontent.com/assets/17108017/18471754/a5022438-79b3-11e6-8814-f6138b3cb22b.jpg)

lucaderi:
@Niko78 Can you please check on the nProbe side how many flows are sent to nProbe? I suggest you to add -b 1 (or even -b 2) on nProbe and see. I see flow drops on ntopng: can you please create a virtual interface per exporter router/nProbe so that you can leverage on the multithreading ntopng capabilities?

Niko78:
I hope it is what you wanted, I let nprobe launched during 1 minute

13/Sep/2016 14:30:10 [nprobe.c:2718] Processed packets: 4414 (max bucket search: 0)
13/Sep/2016 14:30:10 [nprobe.c:2701] Fragment queue length: 0
13/Sep/2016 14:30:10 [nprobe.c:2727] Flow export stats: [2166136 bytes/4409 pkts][102 flows/4 pkts sent]
13/Sep/2016 14:30:10 [nprobe.c:2737] Flow drop stats:   [0 bytes/0 pkts][0 flows]
[...]

PR #750, opened 2016-09-16
Fix user activity RRA sizes, make RRA preferences configurable and disable activity log by default

emanuele-f:
#740

### DIFF
--- a/scripts/lua/modules/graph_utils.lua
+++ b/scripts/lua/modules/graph_utils.lua
@@ -932,20 +932,21 @@ end
 
 -- ########################################################
 
-function createActivityRRDCounter(path, step, verbose)
+function createActivityRRDCounter(path, verbose)
    if(not(ntop.exists(path))) then
       if(verbose) then io.write('Creating RRD '..path..'\n') end
       local prefs = ntop.getPrefs()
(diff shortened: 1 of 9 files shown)


scripts/lua/modules/graph_utils.lua
```diff
@@ -932,20 +932,21 @@ end
...
-      local hb = step * 2 -- Default hb = 2 minutes
+      local step = 300
+      local hb = step * 2
       ntop.rrd_create(
 	 path,
 	 step,
 	 'DS:in:DERIVE:'..hb..':U:U',
 	 'DS:out:DERIVE:'..hb..':U:U',
 	 'DS:bg:DERIVE:'..hb..':U:U',
-         'RRA:AVERAGE:0.5:1:'..tostring(prefs.other_rrd_raw_days*24*(3600/step)),
```
lucaderi, line 946: This change has nothing to do with this bug. I think step was not a bad idea.

emanuele-f: I've changed pref from days to hours to possibly reduce RRA sizes. Now it only takes 196K per host (versus old 3 MB).

Issue #731, opened 2016-09-07
Feature Request: interface with sample rate


blues-git:
I have few devices that I'm monitoring using ntopng.
Traffic is directed using port mirror with sample-rate 1:100. That causes bandwidth problems (100x lower rates).

I would be great to introduce option to specify interface with sample-rate, i.e.: -i eth0@100

emanuele-f:
Implemented as a UI interface setting. Thank you for the suggestion.

blues-git:
Great.
Version 2.5.161028-1751 from debian packages works fine.
Thanks.

PR #799, opened 2016-10-26
Specify pcap interface sampling rate to scale stats

emanuele-f:
For #731. To be reviewed before merge. Note: if_stats.lua layout change.


emanuele-f:
Implemented in 50db46c and 0c917a4

PR #819, opened 2016-11-12
request a network historical research

Niko78:
changed getNumFlows function to get it working with network entry (ie 192.168.155.0/24)
I tried on my own ntopng and it works well


simonemainardi:
thanks for providing the patch. However, it is not really useful per-se. Indeed, it just gives you the count of flows for an ipv4 subnet but then you won't be able to browse those flows. Please, see if you can extend, at least, pages that allows to navigate IPv4 and IPv4 flows as well.

Niko78:
for me it works as expected
I can see all my flows after clicking ipv4 tab

PR #820, opened 2016-11-13
Do not use clang just because it is installed somewhere

fmeerkoetter:
Just picking up /usr/bin/clang++ breaks for example Yocto
cross compiles if clang++ is installed on the build host.

Instead use AC_PROG_CXX to detect a C++ compiler.
If clang should be used set CXX on the shell to clang++ before
running configure.


lucaderi:
Thank you

PR #821, opened 2016-11-14
Use AC_CHECK_HEADER to test for hiredis

fmeerkoetter:
Testing for /usr/include/hiredis is bad for cross compiles.

This commit fixes also the problem that -I/usr/include/hiredis
was added twice to HIREDIS_INC. Once manually and once through
pkg-config --clags.

### DIFF
--- a/configure.seed
+++ b/configure.seed
@@ -160,13 +160,15 @@ fi
 
 DATE=`date +"%Y-%m-%d"`
 
-if test -d "/usr/include/hiredis"; then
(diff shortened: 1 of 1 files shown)


lucaderi:
Please update your pull request

fmeerkoetter:
I will. Have to find the time to figure out why it works here locally, while it fails in the CI.

lucaderi:
Please resubmit a patch when working with the current code release

Issue #835, opened 2016-11-27
IPv4/IPv6 search


lucaderi:
When a host has both an IPv4 and an IPv6 address pointing to the same symbolic name, searches are difficult.
- In the top right field, there are two items (one for IPv4 and one for IPv6) with the same symbolic name, but nowhere it is mentioned which host has the IPv4 and which the IPv6 address
- The search result is http://localhost:3000/lua/host_details.lua?host=SYMBOLIC and as both IPv4 and v6 addresses share the same symbolic name only one host can be found.

PR #836, opened 2016-11-29
Update typeahead script and use ip addresses when searching hosts

emanuele-f:
The migration of existing code to avoid hostname-based access to the host_details page should be gradual since it is spread across different scripts in the codebase. This patch only fixes search bars related to #835.

### DIFF
--- a/scripts/lua/find_host.lua
+++ b/scripts/lua/find_host.lua
@@ -25,10 +25,30 @@ print [[
       res = interface.findHost(query)
 
       if(res ~= nil) then
+	 values = {}
 	 for k, v in pairs(res) do
-	    if(v ~= "") then 
+	    if(v ~= "") then
+	       if not values[v] then
+	          values[v] = 1
+	       else
+	          values[v] = values[v] + 1
+	       end
+	    end
+	 end
+	 
+	 for k, v in pairs(res) do
+	    if(v ~= "") then
+	       if values[v] > 1 then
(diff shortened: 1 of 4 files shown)


scripts/lua/find_host.lua
```diff
@@ -25,10 +25,30 @@ print [[
...
+	          values[v] = values[v] + 1
+	       end
+	    end
+	 end
+	 
+	 for k, v in pairs(res) do
+	    if(v ~= "") then
+	       if values[v] > 1 then
+	          -- we matched both an ipv4 and ipv6 with same host name, display differently
+	          if findString(v, ":") == nil then
```
simonemainardi: check lua_utils there are already functions that check if the IP is v4 or v6 and they should also be more accurate

simonemainardi (changes requested):

scripts/lua/find_host.lua
```diff
@@ -35,15 +35,15 @@ print [[
 	       end
 	    end
 	 end
-	 
+	 tprint(res)
```
simonemainardi: remove this debug line

simonemainardi (approved):

simonemainardi:
thank you @emanuele-f

lucaderi:
@emanuele-f @simonemainardi This code is bugged and never tested. See " if isIPv6String(v) then" in find_host.lua and you will understand why it doesn't work.

PR #837, opened 2016-11-29
Traffic policing page i18n, are-you-sure integration and styling

emanuele-f:


scripts/lua/if_stats.lua
```diff
@@ -19,6 +19,12 @@ require "graph_utils"
 require "alert_utils"
 require "db_utils"
 
+local i18n = require "i18n"
```
simonemainardi: localization is done only one time in lua utils so you should avoid repeating the same code: https://github.com/ntop/ntopng/blob/dev/scripts/lua/modules/lua_utils.lua#L11

scripts/lua/if_stats.lua
```diff
@@ -1321,7 +1325,7 @@ $("#network").change(function() {
 
 local shapers = shaper_utils.getSortedShapers(ifid)
 
-function print_network_shapers(shapers, curshaper_id, terminator)
+function print_shapers(shapers, curshaper_id, terminator)
```
simonemainardi, line 1322: should this method be placed in shaper_utils?

scripts/locales/en.lua
```diff
@@ -12,6 +12,16 @@ local  en = {
    difference = "Difference",
    total = "Total",
    today = "Today",
+   NOTES = "NOTES",
```
simonemainardi: why all caps? make it lowercase as all the other fields

scripts/lua/if_stats.lua
```diff
@@ -1355,8 +1359,9 @@ print[[
 <script>
 ]] print(jsFormCSRF('deleteNetworkForm', true)) print[[
 ]] print(jsFormCSRF('editNetworksForm', true)) print[[
-function doAddDeleteButton(td_idx, callback_str) {
-   $("td:nth-child("+td_idx+")", $(this)).html('<a href="javascript:void(0)" class="add-on" onclick="' + callback_str + '" role="button"><span class="label label-danger">Delete</span></a>');
+function doAddDeleteButton(td_idx, callback_str, label) {
```
simonemainardi: redundant word "do" in method name.
Same of many other methods.

scripts/lua/if_stats.lua
```diff
@@ -1960,6 +2014,39 @@ print [[</form>
    var hash = window.location.hash;
    $('#filterPageTabPanel a[href="' + hash + '"]').tab('show');
    /*** End Page Tab State ***/
+   $(function() {
```
simonemainardi: see if you can generalize this method to use it both here and for the preferences page

simonemainardi (changes requested):

scripts/lua/if_stats.lua
emanuele-f: I use it as a reminder for callbacks, usually invoked after some user action. Is this bad?

scripts/locales/en.lua
emanuele-f: maybe we will add a "Notes" string in the future (only first letter capitalized), so I want to distinguish this here. As a workaround, I can place a notes -> "NOTES" entry under "shaping"

scripts/lua/if_stats.lua
simonemainardi: how would you expect that people know your conventions? if it's a callback, call it cbBlaBla, callbackBlaBla something that is intuitive

scripts/lua/if_stats.lua
emanuele-f: It's really visualization bound, I wouldn't put it into a util file. In future, all the policing related code should be moved entirely to the pro side

simonemainardi (approved):

simonemainardi:
thank you @emanuele-f

PR #853, opened 2016-12-10
Flow alerts explorer integration and LUA alerts query unification

emanuele-f:
Please note that *pro* part of this work is needed


simonemainardi:
thanks, I've already merged the changes.

Remember to fix the alert deletion when Type or Severity is selected. That is, if I delete the latest alert with a given severity, the tab disappears.

PR #861, opened 2016-12-19
Traffic profiles page improvementes

emanuele-f:


emanuele-f:
Merged with @830ca723 and previous commits. Thank you.

PR #880, opened 2016-12-29
Implement resolution buttons in the preferences page

emanuele-f:
Please check the min and max values where "TODO check min/max" is specified


simonemainardi:
empty values are allowed and saved to redis as NaNs

image

```
127.0.0.1:6379> get ntopng.prefs.local_host_max_idle
"NaN"
```

prevent this behavior

emanuele-f:
According to variable names in code, it seems like empty fields should restore default value, but this seems broken since the default parameter passed is not really the default but the previous value of the field. What's the intented behaviour?

simonemainardi:
prevent empty fields from being saved

simonemainardi:
image

simonemainardi:
merged manually after test and extra works. Thank you.

PR #881, opened 2016-12-29
Column alignment, drilldown, calendar button change

emanuele-f:
Requires pro branch.


simonemainardi:
text field contents are cut even if there's enough space on the sides

image

reducing page width yields

image

lucaderi:
This patch changed the alignment, so I accept it. But the problem of the calendar is not fixed (no JS errors).

lucaderi:
Please submit a new pull request that addresses the remaining issues

PR #886, opened 2016-12-30
Fix the bug in function timeval_diff.

kldeng:
When the tv_sec of result is an odd number, we should add 0.5sec to
tv_usec of result after the tv_sec was divided by 2.


simonemainardi:
thanks, I've re-coded it in a more compact way: https://github.com/ntop/ntopng/commit/3c2b020421d6e1f5be204efb129f4a0e2f33d6a1

thank you!

PR #905, opened 2017-01-05
Shapers improve

emanuele-f:
Requires pro branch.
C side categories support required.


simonemainardi:
@emanuele-f can you please resolve the conflict?

PR #918, opened 2017-01-10
Prefs page layout and log level fix

emanuele-f:


lucaderi:
The main problems are now fixed but you need to improve the layout still a bit.

Numeric fields should be smaller in size (contrary to string fields that need to be longer) and they need to be hooked to the buttons so they stay on the same row at anytime.
screen shot 2017-01-10 at 05 16 12

emanuele-f:
Here are input sizes after the change:

screenshot at 2017-01-10 16-45-38

screenshot at 2017-01-10 16-45-21

For the row split issue, I regret to say it seems to be damn difficult to make the layout behave that way. I suspect that mixing html tables and styles with bootstrap classes is one of problems.

lucaderi:
Looks almost good now. Cherry pie:
- please reduce the space between the buttons and the text field
- in case a non-numeric value is put onto the numeric field, is it possible to avoid non numeric chars at all or the error message you report in red is the best that you can do?

emanuele-f:
- I've finally managed to get a fixed distance between buttons and text field, please check out. Fields now stay in line too!
- The error message is shown by a jquery validator, which does not support that feature. Using the stock html validator actually prevents inserting numbers (in chrome only), but it also leads to the graphic issue in the picture. Since browser support is poor, I think this is not possible.

screenshot at 2017-01-10 19-52-30

PR #920, opened 2017-01-10
Create export.lua to bypass csrf token

kuljule:
This allows users to curl or wget the bandwidth data so that they can download it onto their computer.

### DIFF
--- a/export.lua
+++ b/export.lua
@@ -1,0 +1,57 @@
+--
+-- (C) 2013-16 - ntop.org
+--
+
+-- This lua script is basically the same as do_export_data.lua except this bypasses the csrf token to allow you to export the data without the csrf token.
+-- Put this script in pfsense directory /usr/local/share/ntopng/scripts/lua/ to allow access to the data through curl.
+
+dirs = ntop.getDirs()
+package.path = dirs.installdir .. "/scripts/lua/modules/?.lua;" .. package.path
+
+require "lua_utils"
+
+sendHTTPHeader('application/json')
+
+-- Only needed to change ~= nil to == nil.
+
+if(_GET["csrf"] == nil) then
+
+interface.select(ifname)
+if((_GET["hostIP"] ~= nil) and (_GET["hostIP"] ~= "")) then
+   vlan = 0
+   if ((_GET["hostVlan"] ~= nil) and (_GET["hostIP"] ~= "")) then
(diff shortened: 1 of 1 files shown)


simonemainardi:
thanks for submitting the code, however, there are several things that doesn't allow us to accept the patch:

- bypassing CSRF checks is not the right strategy and there's not even the need to perform CSRF checks unless you submit a form
- you don't select the interface (`ifname`) and just take the default one
- `_GET["hostIP"]` and `_GET["hostVlan"]` are not standard, you should use the helper function `url2hostinfo` that parses host ip and vlan
- indentation is broken
- hosts info is retrieved twice in `getHostsInfo` and then in `getHostInfo`
- library `dkjson` should be used to generate json replies.

That said, ntopng already supports what you did in `export.lua`:
1. to retrieve JSON information for a single host you can query this endpoint: `http://localhost:3000/lua/host_get_json.lua?ifId=1&host=192.168.1.111`
2. to retrieve JSON information for all hosts you can query `http://localhost:3000/lua/get_hosts_data.lua?mode=all&currentPage=1&perPage=10&sortColumn=column_thpt&sortOrder=desc` to get a JSON list of hosts and then you can perform 1. for every item.

kuljule:
Thank you so much for your response. I'm not suggesting changing anything that has already been created. I'm suggesting adding this script for people that want to export all of the ntopng data for users so that they can find whatever they want to look for. 

This is simply the same script as do_export_data.lua except I changed 
`if(_GET["csrf"] ~= nil) then` to `if(_GET["csrf"] == nil) then`

As far as your steps to do what I'm doing in export.lua.... I know that you can find every ip address that is on your network and then do what you did for 1. But that would be pretty hard with a network of like 200 users. What's good about export.lua is that it gets all of that information in 1 script so that you can easily take that, convert it to a csv file and then easily import it into a mysql database.

simonemainardi:
@kuljule we can consider the inclusion of this script but please first you should address the points discussed above.

simonemainardi:
please, consider the suggested changes and resubmit. Thank you

kuljule:
The suggested changes should also be considered for do_export_data.lua as well then because my code is basically a copy of that. 

Thanks

kuljule:
Also please give examples of each of your points. I don't really understand what you are asking for in them. 

Thanks.

PR #922, opened 2017-01-12
Fix filters being lost after selecting local/remote hosts filter

emanuele-f:


simonemainardi:
merged after review and test. thank you.

PR #923, opened 2017-01-12
Refresh blacklist rules when blacklist alert preference changes

emanuele-f:


simonemainardi:
merged after manual rebase. thank you

PR #929, opened 2017-01-16
Create a preference to enable top sites

emanuele-f:
Calculating top sites in redis can be disk-intensive task for big networks.
This is now disabled by default and can be manually enabled from preferences page.


simonemainardi:
merged manually after rebase, thank you.

PR #951, opened 2017-01-25
Replace getFlowPeers function with getFlows and limit results to reduce memory footprint

emanuele-f:
Other function to be checked: getHostsInfo, limit it's too high.
Note: please check host_details.lua, page "peers" sometimes is inaccurate!


simonemainardi:
accepted after manual rebase and fixes. Thank you

PR #972, opened 2017-02-08
Select host pool in host_details page

emanuele-f:
screenshot at 2017-02-08 12-41-29

### DIFF
--- a/scripts/lua/modules/lua_utils.lua
+++ b/scripts/lua/modules/lua_utils.lua
@@ -980,6 +980,18 @@ function isValidPoolMember(member)
   return false
 end
 
+function host2member(ip, vlan)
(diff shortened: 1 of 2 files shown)


scripts/lua/modules/lua_utils.lua
```diff
@@ -980,6 +980,18 @@ function isValidPoolMember(member)
   return false
 end
 
+function host2member(ip, vlan)
```
simonemainardi, line 983: see function hostinfo2hostkey, I think you can reuse it, e.g., hostinfo2hostkey({host="",vlan=0}) perhaps you can extend it with subnet

emanuele-f:
Manually merged.

PR #974, opened 2017-02-10
Adjustments to obtain an OpenBSD build

numeronove:
Attempting to build OpenBSD fails for various reasons.

First and foremost, there is a number of packages that have to
be installed as build dependencies

Then, small updates are needed to the autotools files to have autogen.sh
and configure, to exit without errors.

doc/README.OpenBSD describes the whole process in fair detail.

Last but not least, there are type errors that g++ compiler catches in
OpenBSD but not in Linux. To have the build, a few source files
were updated to have effect only in OpenBSD, leaving other operating
systems untouched.

With only one exception, a prototype in include/Flow.h was updated to
stay in sync with the implementation.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -1,6 +1,7 @@
 @PRO_MAKEFILE_INC@
 prefix = @prefix@
 datadir = @datadir@
+datarootdir = @datarootdir@
 SHELL=/bin/sh
 OS := $(shell uname -s)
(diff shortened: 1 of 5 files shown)


lucaderi:
Thank you

numeronove:
You're welcome :-)

simonemainardi:
thank you @numeronove !

Issue #581, opened 2016-05-20
Implement Websocket for AJAX refresh


lucaderi:
ntopng must start using web sockets at least for the page footer to reduce the load on the browser

tomlabaude:
+1
Client opens 1 socket per second simply for network_load.lua, a wireshark trace is full of TCP handshakes.
Reusing the same socket will save files opened on the client, reduce load on the network, on the server and will more ... pretty!

<img width="1088" alt="screenshot" src="https://cloud.githubusercontent.com/assets/15690818/19184474/9a8d4e74-8c7d-11e6-9feb-06f32b109d4a.png">

lucaderi:
@tomlabaude Would you like to lead this development ?

tomlabaude:
Unfortunately, I'm not able to do that dev and am also looking for a dev to work on the frontend of another project.
I have mixed feelings of only being able to report problems without being able to help...

lucaderi:
For the time being we decided that websocket support can create issues to the HTTP server and thus we postpone its implementation.

PR #986, opened 2017-02-16
Websockets

emanuele-f:
Implements websockets support in ntopng #581.
Uses websockets for network_load.lua and hosts_stats.lua.

### DIFF
--- a/src/HTTPserver.cpp
+++ b/src/HTTPserver.cpp
@@ -824,7 +824,7 @@ HTTPserver::HTTPserver(const char *_docs_dir, const char *_scripts_dir) {
 
   limiter = new HTTPlimiter();
 
-  snprintf(num_threads, sizeof(num_threads), "%d", HTTP_SERVER_NUM_THREADS);
+  snprintf(num_threads, sizeof(num_threads), "%u", Utils::getNumHTTPServerThreads());
 
   static char *http_options[] = {
     (char*)"listening_ports", ports,
@@ -940,6 +940,13 @@ HTTPserver::HTTPserver(const char *_docs_dir, const char *_scripts_dir) {
   /* ***************************** */
 
   ntop->getTrace()->traceEvent(TRACE_NORMAL, "Web server dirs [%s][%s]", docs_dir, scripts_dir);
+  ntop->getTrace()->traceEvent(TRACE_NORMAL,
+			       "Web server threads: %u servers, "
+			       "%u websocket servers, "
+			       "%u maximum servers per host",
+			       Utils::getNumHTTPServerThreads(),
+			       Utils::getNumHTTPWebSocketServerThreads(),
+			       Utils::getMaxNumHTTPServerThreadsPerHost());
 
   if(use_http)
     ntop->getTrace()->traceEvent(TRACE_NORMAL, "HTTP server listening on port(s) %s", 
--- a/src/HTTPserver.cpp
+++ b/src/HTTPserver.cpp
@@ -618,12 +618,6 @@ static int handle_lua_request(struct mg_connection *conn) {
   if ((mg_get_header(conn, "Upgrade") != NULL) && (strcmp(mg_get_header(conn, "Upgrade"), "websocket") == 0))
     is_websocket_request = true;
 
-  /* Limit connections */
-  if (! httpserver->get_connections_limiter()->connectHost(request_info->remote_ip, request_info->remote_port, is_websocket_request)) {
-    return send_error(conn, 503 /* Service Unavailable */, request_info->uri,
-      "Too many concurrent requests");
-  }
-
   if((strncmp(request_info->uri, "/lua/", 5) == 0)
      || (strcmp(request_info->uri, "/") == 0)) {
     /* Lua Script */
@@ -749,14 +743,19 @@ static int handle_websocket_connect(const struct mg_connection *conn) {
 /*
  * This is called when the websocket becomes ready. Application can send data.
  */
-static void handle_websocket_ready(struct mg_connection *conn) {
+static int handle_websocket_ready(struct mg_connection *conn) {
   struct mg_request_info *request_info = mg_get_request_info(conn);
   websocket_user_data *user_data = (websocket_user_data *)request_info->user_data;
 
   if (ntopng_is_busy())
-    return;
+    return(1);
+
+  /* Limit WebSocket connections */
+  if (! httpserver->get_connections_limiter()->connectHost(request_info->remote_ip, request_info->remote_port, true))
+    return(1);
 
   user_data->lua->handle_websocket_ready(user_data->script_path);
+  return (0);
 }
 
 /*
--- a/src/HTTPserver.cpp
+++ b/src/HTTPserver.cpp
@@ -481,6 +481,44 @@ static void uri_encode(const char *src, char *dst, u_int dst_len) {
 
 /* ****************************************** */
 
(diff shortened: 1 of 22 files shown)


src/HTTPserver.cpp
```diff
@@ -35,6 +35,101 @@ static HTTPserver *httpserver;
...
+
+/* The maximum number of threads to dedicate do WebSocket connections */
+#define SERVER_MAX_WEBSOCKET_THREADS 4
+
+/* The maximum number of active connections (busy threads) per host */
+#define SERVER_MAX_THREADS_PER_HOST 8
+
+/* ****************************************** */
+
+class ConnectionsLimiter {
```
simonemainardi: please use standard Header (and possibly cpp) file to declare/define this class.

src/HTTPserver.cpp
```diff
@@ -35,6 +35,101 @@ static HTTPserver *httpserver;
 
 /* ****************************************** */
 
+/* The total number of threads to use in mongoose */
+#define SERVER_NUM_THREADS 16
```
simonemainardi: use ntop_defines.h for these constants

src/HTTPserver.cpp
```diff
@@ -35,6 +35,101 @@ static HTTPserver *httpserver;
...
+
+/* ****************************************** */
+
+class ConnectionsLimiter {
+  private:
+    Mutex *mutex;
+    int active_threads;
+    int active_websockets;
+    struct {
+      uint32_t ip;
```
simonemainardi: what happens with ipv6?

src/HTTPserver.cpp
```diff
@@ -35,6 +35,101 @@ static HTTPserver *httpserver;
...
+      active_threads = 0;
+      active_websockets = 0;
+      memset(active_hosts, 0, sizeof(active_hosts));
+    }
+
+    ~ConnectionsLimiter() {
+      delete mutex;
+    }
+
+    bool connectHost(uint32_t ip, int port, bool is_websocket) {
```
simonemainardi: handle ipv6 not only ipv4

src/HTTPserver.cpp
```diff
@@ -35,6 +35,101 @@ static HTTPserver *httpserver;
...
+    }
+
+    void disconnectHost(uint32_t ip, int port) {
+      mutex->lock(__FILE__, __LINE__);
+
+      for (int i=0; i<active_threads; i++) {
+        if ((active_hosts[i].ip == ip) && (active_hosts[i].port == port)) {
+          bool is_websocket = active_hosts[i].websocket;
+
+          memmove(&active_hosts[i], &active_hosts[i+1], (active_threads - i - 1) * sizeof(active_hosts[0]));
```
simonemainardi: if active_threads == SERVER_NUM_THREADS you access an invalid memory location with &active_hosts[i+1]

src/HTTPserver.cpp
```diff
@@ -35,6 +35,101 @@ static HTTPserver *httpserver;
...
+          memset(&active_hosts[--active_threads], 0, sizeof(active_hosts[0]));
```
simonemainardi: zeroing this is not needed as you use active_threads to access active_hosts

src/HTTPserver.cpp
```diff
@@ -649,6 +783,131 @@ static int handle_lua_request(struct mg_connection *conn) {
...
+ */
+static int handle_websocket_data(struct mg_connection *conn, int bits, char *data, size_t data_len) {
+  int rc = 0;
+  struct mg_request_info *request_info = mg_get_request_info(conn);
+  websocket_user_data *user_data = (websocket_user_data *)request_info->user_data;
+
+  if (ntopng_is_busy())
+    return(1);
+
+  char *str_data = (char*) malloc(data_len + 1);
```
simonemainardi, line 774: why you don't use strdup?

simonemainardi (changes requested):
please review suggested comments, add a short README.websocket for the developer that will use it, and then feel free to merge

src/HTTPserver.cpp
emanuele-f: It does not happen since the expression `&active_hosts[i+1]` just computes an address and in the case you spotted `active_threads - i - 1 = 0` so no memory move takes place.

src/HTTPserver.cpp
emanuele-f: Because input parameter `(char*) data` is not guaranteed to be null terminated

src/HTTPserver.cpp
simonemainardi: ISO/IEC 9899:1999:

$7.20.5
>  These utilities make use of a comparison function to search or sort arrays of unspecified type. Where an argument declared as size_t nmemb specifies the length of the array for a function, nmemb can have the value zero on a call to that function; [...] Pointer arguments on such a call shall still have valid values, as described in 7.1.4.

7.1.4
> If an argument to a function has an invalid value (such as a value outside the domain of the function, or a pointer outside the address space of the program, or a null pointer, or a pointer to non-modifiable storage when the corresponding parameter is not const-qualified) or a type (after promotion) not expected by a function with variable number of arguments, the behavior is undefined.

So even if a number of bytes equal to zero is an acceptable value, the out-of-bounds pointer is not.

src/HTTPserver.cpp
simonemainardi: see `strndup`

src/HTTPserver.cpp
emanuele-f: I appreciate some help to get the ipv6 address out of mongoose.c

src/HTTPserver.cpp
emanuele-f: That's an interesting point. Standard is quite vague about this explicit case, but it's worth fixing it

simonemainardi:
please review the implementation as there's plenty of errors in every page:

image

image

emanuele-f:
All the errors are related to the commit a0fdfec .

Note: only allowing 2 HTTP threads (2 are ws dedicated) per host is really limitative since browsers make requests concurrently. I think binding the number of HTTP threads to the hardware concurrency in this way is not a good idea.

simonemainardi:
> All the errors are related to the commit a0fdfec .

Commit makes a more conservative use of resources by taking into account the cores available to limit the number of server threads. Otherwise you would spawn 16 threads also on a raspberry pi and I am not sure this is ideal. Thus, if limiting the number of threads causes 503 errors, then there's something that should be re-discussed in the design. Multiple requests can be naturally queued and served by the same pool of threads.

lucaderi:
@simonemainardi @emanuele-f Can you please fix this issue and incorporate the change?

emanuele-f:
So, let's decide if we want to integrate this feature.

Pros:
- all the footer updates will go through the websocket, reducing a litthe network overhead and response time (of the footer)

Cons:
- some of the threads are dedicated to the ws channel, which will reduce the total amount of concurrent requests available (this can be mitigated as described below)
- the websocket code needs further security review, it is based upon an old mongoose code

What I feel is really missing here is the possibility to route *all* the ajax calls through the websocket, in a transparent way (populating the _GET parameters of the scripts from the ws buffer). *That* would really reduce the network overhead. @simonemainardi and @lucaderi what do you think about it?

lucaderi:
For the time being we decided that websocket support can create issues to the HTTP server and thus we postpone its implementation.

Issue #1001, opened 2017-02-25
Split IPv4/v6 View


lucaderi:
In both flows and hosts view, on the table drop-down menu, add the ability to select only IPv4 or IPv6 traffic only (better if it can be combined with local IPv4, remote IPv6 etc).

PR #1007, opened 2017-02-28
More flows and hosts filters

emanuele-f:
Implements #1001

- Adds a filter for local/remote hosts for flows (one for client, one for server)
- Adds ip version filter for both flows and hosts
- Fixes "Unknown" protocol filter not being applied


simonemainardi:
accepting, please review suggested changes directly on dev

PR #1017, opened 2017-03-05
Spelling

jsoref:


httpdocs/js/sequence_sunburst.js
```diff
@@ -69,7 +69,7 @@ function SequenceSunburst(circle_name,sequence_name,refresh,update_url,url_param
   ///////////////////////////////////////////////////////////
   // UPDATE FUNCIONTS ////////////////////////////////////
   ///////////////////////////////////////////////////////////
-  this.falshUpdate = function (start,end) {
+  this.flashUpdate = function (start,end) {
```
jsoref, line 72: I'm not sure about this change

simonemainardi:
hi, thanks for submitting the code, that is a very accurate job. However, it fails to build, please fix it so that we can approve. (See tokenizer).

simonemainardi:
manually merged after fix, rebase and squash (https://github.com/ntop/ntopng/commit/78e78c018cdb6a7e979be1a828159c8de0b8ad8d). Thank you.

src/Flow.cpp
```diff
@@ -1709,15 +1709,15 @@ json_object* Flow::flow2json() {
 
   if(json_info && strcmp(json_info, "{}")) {
     json_object *o;
-    enum json_tokener_error jerr = json_tokener_success;
+    enum json_tokenizer_error jerr = json_tokenizer_success;
```
jsoref, line 1712: I did worry that this was not local. Fixing this would involve changing the upstream module (if it's willing to do that) and then updating the downstream. I need to figure out where your upstream modules live.

Thanks for dropping it.

Issue #1144, opened 2017-04-08
Logstash errors when parsing data from ntopng


jevonearth:
Hi,

I have been testing the ntopng to logstash feature, and I'm getting some good data passed from: mikrotik -> nprobe -> ntopng -> logstash -> elasticsearch

But I get regular json parse errors where there are spurious characters (such as `\u00004` and `\u0000` ) in the json. See two log samples below, along with version information and copies of my configs. 

Please let me know if I can test anything specific in order to help you further. 

Thanks,
-Jev

## Logstash errors on stdout:
```[2017-04-08T18:41:10,831][WARN ][logstash.codecs.jsonlines] JSON parse error, original data now in message field {:error=>#<LogStash::Json::ParserError: Illegal character ((CTRL-CHAR, code 0)): only regular white space (\r, \n, \t) is allowed between tokens
 at [Source: 4, "FIRST_SWITCHED": 1491676739, "LAST_SWITCHED": 1491676739, "json": { "15": "184.70.24.213", "10": "2", "14": "12", "5": "0", "16": "15169", "17": "0" }, "CLIENT_NW_LATENCY_MS": 0.000000, "SERVER_NW_LATENCY_MS": 0.000000, "DST_IP_COUNTRY": "US", "DST_IP_LOCATION": [ -122.057404, 37.419201 ], "NTOPNG_INSTANCE_NAME": "ntop-redder-dev", "INTERFACE": "tcp:\/\/127.0.0.1:5559c [probeIP: 10.60.58.1]" }; line: 1, column: 2]>, :data=>"\u00004, \"FIRST_SWITCHED\": 1491676739, \"LAST_SWITCHED\": 1491676739, \"json\": { \"15\": \"184.70.24.213\", \"10\": \"2\", \"14\": \"12\", \"5\": \"0\", \"16\": \"15169\", \"17\": \"0\" }, \"CLIENT_NW_LATENCY_MS\": 0.000000, \"SERVER_NW_LATENCY_MS\": 0.000000, \"DST_IP_COUNTRY\": \"US\", \"DST_IP_LOCATION\": [ -122.057404, 37.419201 ], \"NTOPNG_INSTANCE_NAME\": \"ntop-redder-dev\", \"INTERFACE\": \"tcp:\\/\\/127.0.0.1:5559c [probeIP: 10.60.58.1]\" }"}
```
```
[...]

PR #1147, opened 2017-04-11
Code revision for issue #1144

bperian:
* Fix for https://github.com/ntop/ntopng/issues/1144

 **Changes done**
* Sending on socket is done on the actual data size (not sending full buffer)
* Added timestamp for Logstash export 
* Modified Logstash example configuration to use the ntopng timestamp for flows


doc/README.logstash
```diff
@@ -35,6 +34,11 @@ input {
 
 filter {
    if [type] == "ntopng-ls" {
+     date {
+                match => ["ntop_timestamp" , "yyyy-MM-dd'T'HH:mm:ss.SSSZ"]
```
simonemainardi, line 38: @bperian  can you please explain why you need this apparently double timestamp?

src/Flow.cpp
```diff
@@ -1700,16 +1700,28 @@ json_object* Flow::flow2json() {
...
     tm_info = gmtime(&t);
 
     strftime(buf, sizeof(buf), "%FT%T.0Z", tm_info);
-    json_object_object_add(my_object, "@timestamp", json_object_new_string(buf));
+
+    if(ntop->getPrefs()->do_dump_flows_on_ls()){
+      /*  Add current timestamp differently for Logstash, in case of delay 
+       *  Note: Logstash generates it's own @timestamp field on input
+       */
+      json_object_object_add(my_object,"ntop_timestamp",json_object_new_string(buf));
```
simonemainardi, line 1740: can you re-use the `@timestamp` used for elastic search?

simonemainardi (changes requested):

src/Flow.cpp
bperian: There will be a delay between the actual insert into Elasticsearch by default so the timestamp will be different. If we send the original timestamp and replace the Logstash generated value of @timestamp the data will be mapped better in visualizers. 
 We can delete the ntop_timestamp field afterwards in logstash so we don't have an extra field.

bperian: I tested with reusing the `@timestamp `already defined in ntop, but Logstash overwrites it by default.

bperian:
@simonemainardi 

 No, by default JSON values are overriden by logstash if the field names are identical so in order to get the correct timestamp in the right field the date and mutate are required

simonemainardi (approved):

simonemainardi:
@bperian thanks. Please fix the conflicts and I'll approve.

simonemainardi:
thank you

PR #1150, opened 2017-04-12
Prevent assigning a network to the multiple Host Pools

### LINKED ISSUE #1130: Adding subnet to different pools
You can add the same subnet to different pools. This should not be possible. 
What policy do you apply in this case?

emanuele-f:
Enforces MAC/Host/Network consistency across Host Pools. Fixes #1130.

- Networks are now normalized with their network prefix before being dumped to Redis
- No network duplication is now possible inside the same pool or between different pools

Note: currently, checks are currently implemented Lua side. No javascript / ajax errors are shown when an addPoolMember operation fails.

### DIFF
--- a/include/HostPools.h
+++ b/include/HostPools.h
@@ -26,6 +26,7 @@
 
 class NetworkInterface;
 class Host;
+class Mac;
 #ifdef NTOPNG_PRO
 
 typedef struct {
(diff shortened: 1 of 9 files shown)


simonemainardi:
Please review and apply the following:

* host pool match based on MAC addresses *no longer works*. If I add my NIC MAC address to an host pool, then the host page keeps telling me "Not Assigned".

* the system allows the addition of the same MAC address to different host pools. This must be prevented

* In an host pool, the system doesn't allow the edit of a mac address with an IP address nor with another MAC address. The overwritten mac address is preserved and an extra member with the new value is added.

* Can't delete a MAC address from an host pool. When clicking delete, a dialog appears to ask for confirmation but the the MAC address is not deleted.

* No user notification is received when trying to add one or more already existing networks. The system prevents the addition and this is OK but I would notify the user.

* I noticed that nested addresses are allowed within the same pool. I would allow this behavior as one may remove the least-specific match and leave the more specific match in the same pool.

image

emanuele-f:
Thank you, the reported bugs have been fixed.

> I noticed that nested addresses are allowed within the same pool. I would allow this behavior as one may remove the least-specific match and leave the more specific match in the same pool.

The check for avoiding nested addresses into the same pool is not trivial. For example:
- User adds network A/26
- User adds network A/24

A/24 is not contained in A/26, so the issue cannot be spotted unleass all members combinations are checked (e.g. for each member, create a patricia tree with only the member and check against all other members for inclusion)

simonemainardi:
Thanks for addressing all the issues above. Please, se this and then for me it is OK to merge into dev.


If I try to change the host part of an existing network, the network is removed from the pool

image

image

image

emanuele-f:
Issue fixed. Thank you for your feedback.

simonemainardi:
thank you

Issue #1160, opened 2017-04-14
[Logstash ext] Wrong data sometimes sent to logstash


bperian:
Reported by @jevonearth

If you search the block of text for SERVER_NW_LATEN I think you will narrow down the source. It looks like one block of text is getting overwritten with another.

Possibly issue with how data block is sent .

```
[2017-04-13T19:37:26,665][WARN ][logstash.codecs.jsonlines] JSON parse error, original data now in message field {:error=>#<LogStash::Json::ParserError: Unexpected character ('n' (code 110)): was expecting a colon to separate field name and value
[...]

PR #1161, opened 2017-04-14
Revision for issue #1160 

bperian:
* Wasn't switching buffer position in case of send failure on socket which resulted into malformed JSON


simonemainardi:
thank you @bperian 👍

Issue #1153, opened 2017-04-12
Force admin password change


lucaderi:
For security reasons, it is requested to check (in case the admin user is logged in) if the user has changed the default password, and if not to force the user to change the password to a value other than 'admin'

simonemainardi:
implemented

PR #1197, opened 2017-04-27
Force default admin password change

emanuele-f:
Implements #1153

screenshot at 2017-04-27 21-17-43

### DIFF
--- a/include/Ntop.h
+++ b/include/Ntop.h
@@ -381,6 +381,7 @@ class Ntop {
   bool isInterfaceAllowed(lua_State* vm, int ifid)           const;
   bool checkUserPassword(const char *user, const char *password);
   bool resetUserPassword(char *username, char *old_password, char *new_password);
+  bool mustChangePassword(const char *user);
   bool changeUserRole(char *username, char *user_role) const;
   bool changeAllowedNets(char *username, char *allowed_nets)     const;
   bool changeAllowedIfname(char *username, char *allowed_ifname) const;
--- a/include/ntop_defines.h
+++ b/include/ntop_defines.h
@@ -85,6 +85,7 @@
 
 #define ZMQ_MSG_VERSION           1
 #define LOGIN_URL                 "/lua/login.lua"
+#define LOGOUT_URL                "/lua/logout.lua"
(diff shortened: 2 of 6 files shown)


simonemainardi:
@ValentinaViscarelli can you please check if LDAP auth remains valid with this pull request? 

Thank you

emanuele-f:
Note: the current behaviour with LDAP should be this: if a user selected the LDAP auth but has the default admin password set, then the password change dialog will appear after the LDAP auth takes place.

Note: after being forced to change the default admin password, the user can change it back to the default password from the ntopng user preferences. This is intentional as I think the user should have the possibility to do this.

ValentinaViscarelli:
@simonemainardi and @emanuele-f 
The "password change dialog" appears just for local authentication.

- If you have "local" authentication set then "password change dialog" appears properly.
- If you have "local/ldap" authentication then set the "password change dialog" appears just for local authentication. That means you can authenticate properly with ldap and "password change dialog" doesn't appear.only
- If you have "ldap" authentication set then the "password change dialog" doesn't appear after you have logged into account with ldap.

I don't know that should be the right behavior but this is not true:

_"Note: the current behaviour with LDAP should be this: if a user selected the LDAP auth but has the default admin password set, then the password change dialog will appear after the LDAP auth takes place."_

Let me know if I have to do other tests.

emanuele-f:
@ValentinaViscarelli thank you for your test. But I think the behaviour you get with ldap and local/ldap is not actually correct, can you please try the following steps and report your result?
1) reset redis `redis-cli flushall`
2) login into ntopng and change the password from the dialog
3) go to the users and change admin password back to "admin"
4) go to the preferences and set ldap auth only with your credentials
5) logout from ntopng
6) run `redis-cli del ntopng.prefs.admin_password_changed`.
7) login into ntopng with ldap
8) you should get the password reset dialog appear, is this correct?

ValentinaViscarelli:
@emanuele-f 
Steps from 1 to 7 -> SUCCESS
Step 8 -> FAILED (you don't get password reset dialog but you log on system directly).

emanuele-f:
Got it, thank you for your test.

PR #1203, opened 2017-04-28
LDAP search_path parameter allows spaces

danielmeyer:
Change validation model for LDAP search_path parameter such that criteria may contain spaces.


emanuele-f:
Thank you

PR #1214, opened 2017-05-04
Always show IPv6 suffix in IPv6 host names

emanuele-f:
Requires pro branch (same name)


emanuele-f:
Replaced by commit 9097333cf1e6ff96237c0d86211f545e983cbe6a

PR #1226, opened 2017-05-10
Localize some pages for interface

TakayukiOnodera:


simonemainardi:
thank you @TakayukiOnodera !

scripts/locales/en.lua
```diff
@@ -714,6 +718,11 @@ local  en = {
    },
 
    if_stats_overview = {
+      snmp = "SNMP",
+      iface_switch_active_msg_pre = "The selected interface",
```
emanuele-f, line 722: @TakayukiOnodera please note that in order to localize properly in any language, this kind of messages should be part of a single localization key.
In this example, `iface_switch_active_msg_pre` and `iface_switch_active_msg_after` should be reported as `iface_switch_active_msg = "The selected interface <b>%{interface}</b> [ifid:%{ifid}] is now active"`. 

The key above would be used like this: `print i18n("if_stats_overview.iface_switch_active_msg", {interface=getHumanReadableInterfaceName(ifid), ifid=ifid})`

TakayukiOnodera: @emanuele-f Thank you for pointing that out. I'll correct this kind of messages later on.

emanuele-f: Thank you very much!

PR #1240, opened 2017-05-16
Localize some hosts pages

TakayukiOnodera:


simonemainardi:
thank you @TakayukiOnodera !

PR #1244, opened 2017-05-17
Localize network_details and hosts_stats pages

TakayukiOnodera:
A title of host_stats.lua page was not replaced by single localization key because it depends on a combination of several variables including nil type.


simonemainardi:
Thank you :)

PR #1246, opened 2017-05-19
Localize host details page

TakayukiOnodera:
Host details page is localized except "jaccard" page.

Please note that the code was changed for the messages about blocked traffic in overview page because it's difficult to translate into Japanese for the messages in original code.


emanuele-f:
The blocked traffic message it's ok. Thank you for your effort!

PR #1248, opened 2017-05-21
[bugfix] NTOP LS Export hangs if remote service is disrupted

prutasiscale:
* bug: NTOP LS flow export hangs if remote service gets disrupted 

Fixed code logic to reinitialize connection in case remote service is temporarily disrupted and added warning when NTOP cannot export flows to Logstash

* Tested successfully 

- Start NTOP with LS export with no logstash available , start and stop LS a few times 
- Start NTOP with LS export with logstash available , start and stop LS a few times


bperian:
Should an issue be opened too, for reporting purposes?

simonemainardi:
thank you!

PR #1260, opened 2017-05-24
Quota manual reset and persistency

emanuele-f:
Requires pro branch.

I've performed some tests with iperf, but more tests with real clients should be done.


emanuele-f:
Reset will be handled differently in a simpler way.

PR #1282, opened 2017-06-01
Fix: bubble charts are not displayed

TakayukiOnodera:
Bubble charts for application and L4 protocol between hosts are not displayed. It is solved to pass the detailsLevel="higher" option for getFlowsInfo.


simonemainardi:
Thank you!

PR #1294, opened 2017-06-05
In this FreeBSD conditional add check for libnbpf.a to actually exist.

madpilot78:
Otherwise the build of the OSS version will fail.


simonemainardi:
thank you!

PR #1297, opened 2017-06-06
General fixes

amallia:
* When you allocate an array using the new …[…] syntax, you should deallocate it using delete[]
* Either the condition 'host_ip==NULL' is redundant or there is possible null pointer dereference: host_ip.

### DIFF
--- a/src/Lua.cpp
+++ b/src/Lua.cpp
@@ -97,7 +97,7 @@ static void get_host_vlan_info(char* lua_ip, char** host_ip,
   if(((*host_ip) = strtok_r(buf, "@", &where)) != NULL)
     vlan = strtok_r(NULL, "@", &where);
 
-  if(host_ip == NULL)
+  if(*host_ip == NULL)
     *host_ip = lua_ip;
 
   if(vlan)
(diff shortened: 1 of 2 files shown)


lucaderi:
Thank you Antonio

src/Lua.cpp
```diff
@@ -97,7 +97,7 @@ static void get_host_vlan_info(char* lua_ip, char** host_ip,
   if(((*host_ip) = strtok_r(buf, "@", &where)) != NULL)
     vlan = strtok_r(NULL, "@", &where);
 
-  if(host_ip == NULL)
+  if(*host_ip == NULL)
```
simonemainardi, line 100: this is wrong, host_ip is a pointer to a pointer, and you want to check if it points to a NULL address before assigning to that pointed address.

amallia: @simonemainardi 
Well, the previous behaviour implied checking if NULL and, if so, dereferencing it. This would be a NULL pointer dereference. 
So as i said:

> Either the condition 'host_ip==NULL' is redundant or there is possible null pointer dereference: host_ip

PR #1303, opened 2017-06-08
Localize title of columns on table in hosts_stats

TakayukiOnodera:
dumpFile API call is replaced with code that should be dumped by API.

### DIFF
--- a/scripts/locales/en.lua
+++ b/scripts/locales/en.lua
@@ -93,6 +93,8 @@ local  en = {
    aggregation = "Aggregation",
    refresh = "Refresh",
    l4_protocol = "L4 Protocol",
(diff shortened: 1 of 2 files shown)


emanuele-f:
Thank you!

simonemainardi:
thank yu!,

BTW , can files httpdocs/inc/host_stats_top.inc and httpdocs/inc/host_stats_bottom.inc be removed after this ?

emanuele-f:
Removed in 21d0e55dd0c43ff22a55c153de1aebd2ddb71c4d

TakayukiOnodera:
No, both include files are still referenced from other lua files. I will remove these include files when I have finished to do similar work for other files.

PR #1304, opened 2017-06-08
FreeBSD fetch

madpilot78:
While most linux distributions come with wget preinstalled, FreeBSD does not include it by default. OTOH FreeBSD base system includes a "fetch" command which provides similar functionality.

I modified the Makefile.in to use fetch in place of wget on FreeBSD OS.


simonemainardi:
thank you!

PR #1349, opened 2017-07-05
ntopng init script: missing quotes sometimes cause errors

rbott:
Hi,

I tested the init Script on Debian jessie and sometimes ran into issues with missing quotes. The following fails if $PID is an empty variable:
```
if [ -z $PID ]; then
```
This happens only if there is no pid file at the specified path (or an empty file). The following fixes this:
```
if [ -z "$PID"]; then
```

I also fixed all other cases of missing quotes for parameters/tests in the init script and it works fine on Debian jessie. 

Regards,
Rudolph Bott


simonemainardi:
@ValentinaViscarelli can you review the proposed changes and evaluate them?

ValentinaViscarelli:
@rbott 
Thanks so much!!!

PR #1416, opened 2017-08-29
RPM improvements

gsanchietti:
- Cleanup obsolete tags and scriplets
- Use systemd macros for CentOS 7
- Move systemd unit files to `/usr/lib/systemd/system/n` directory
- Create new ntopng-pcap package without pfring (and dkms) dependencies

`%changelog` sections of both RPMs should be updated before stable release.


simonemainardi:
implemented with https://github.com/ntop/ntopng/commit/b7735213a1a8b5e36efe7238f45e5321c4611b63

PR #1501, opened 2017-10-04
Add support for custom mysql port

andrewmv:
- Fixed port number parsing from startup arguments
  as documented in man page.
- Added port number usage to --help.
- Added port number to error messaging.


simonemainardi:
thanks for submitting but il looks like the port_number is still hard-coded to 3306, that is, it is not parsed from the user submitted @<port>. Please, fix it.

andrewmv:
It's working locally for me, but it looks like I staged my commit incorrectly. I'll re-submit.

andrewmv:
Sorry for the confusion - something's still not quite right with my local repository. I'll update when I've sorted out what's gone wrong.

andrewmv:
I've been unable to locally recreate the build problems with the automated integration tests, so I'm at a loss for how to resolve them, I'm afraid.

andrewmv:
Looks like my build environment was handling C++11 correctly, and the CI tests were not.
Rewrote the logic to be standard of the rest of the project.

simonemainardi:
thanks for submitting!

simonemainardi:
I needed to modify your code as you were leaking memory.

Please see https://github.com/ntop/ntopng/commit/598b9de8432d898f7af9e0a16d70536d159ec35a

The leak was due to the improper use of Utils::tokenizer.

Also, for next contributions, please, consider using proper code indentation and put constants in include/ntop_defines.h rather than hard-coding them.

thank you

PR #1590, opened 2017-12-05
Rename LIBS variable to avoid polluting the environment of submakes l…

madpilot78:
…aunched to compile third party libraries.

While porting the latest version to FreeBSD I had to use the luajit library in the third_parties directory.

It failed to build because hte LIBS variable is polluted by the top level Makefile with the luajit library itself, failing to build the minilua compiler.

This patch fixes it by renaming the LIBS variable in the top level Makefile.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -106,7 +106,7 @@ endif
 
 ######
 TARGET = ntopng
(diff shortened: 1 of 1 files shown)


madpilot78:
This change should be merge to the 3.2-stable branch if accepted.

simonemainardi:
thank you

simonemainardi:
cherry-picked to 3.2-stable as well

PR #1634, opened 2018-01-18
Modify files to allow compiling using llvm/clang 6.0.0 as recently im…

madpilot78:
…ported in FreeBSD 12-CURRENT.

These are brutal modifications I made to make ntopng compile on recent FreeBSD head, after the import of clang 6.0.0.
I made two modifications around.

1) Removed "register" qualifiers. clang has them as deprecated and errors out on them by default now. It states they are incompatible with c++17 standard. I don't think this is a problem. Modern compilers are able to optimise without needing such suggestions.

2) Modified plain calls to bind(2) to explicitly reference bind in the root namespace. If unqualified clang now tries to use the overloaded bind function in the std namespace, which is unrelated.

I can agree that 2 is somewhat brutal and maybe some preprocessor macro trick could be better. I'm open to suggestions on how to make my patch better.

All these changes should also be merged back to the 3.2-stable branch if accepted.

### DIFF
--- a/src/DivertInterface.cpp
+++ b/src/DivertInterface.cpp
@@ -101,7 +101,7 @@ DivertInterface::DivertInterface(const char *name) : NetworkInterface(name) {
   sin.sin_family = AF_INET, sin.sin_port = htons(port);
   sin_len = sizeof(struct sockaddr_in);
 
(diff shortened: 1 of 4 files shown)


simonemainardi:
Squashing and merging after tests. Thank you.

PR #1637, opened 2018-01-23
Additional grafana metric

tobiasdam:
Since internal functions already implement the retrieval of host statistics per category, the metric `allcategories_bps` should be exposed via the grafana api similar to `allprotocols_bps`.


simonemainardi:
thank you

PR #1643, opened 2018-01-25
Separate lsb_release test from PF_RING check for ubuntu packaging

charlesboyo:
When PF_RING is not used, the RELEASE defaults to 'debian' which builds a ntopng package which cannnot be installed Ubuntu 16.04 because it depends on "libhiredis0.10, libmysqlclient18" instead of "libhiredis0.13, libmysqlclient20"


simonemainardi:
thank you

PR #1657, opened 2018-02-15
Redis connection via Unix domain socket

melak:


simonemainardi:
thank you

simonemainardi:
I have fixed the way you determine if the connection is a socket-connection. Indeed, using just as slash was not robust enough. Please check this commit: https://github.com/ntop/ntopng/commit/705425eb0ccd65b7c845d62341f078b11decc5b9

Anyway thanks for your contribution.

Issue #1675, opened 2018-02-26
Throughput is not correctly computed


aouinizied:
Hi all,

This issue is concerning ntopng per flow throughput.
Using zmq export (nprobe running on PC1) exporting to ntopng on PC2. I set nprobe with an avtive timeout of 2 secs in order to refresh intensively stats. 
I start an iperf flow saturating my 1Gbps link.
ntopng shows 164Mbit/s which is uncorrect. However when I compute throughput manually based on nprobe export I get a correct value. 
Moreover, the interface throughput is correct on ntopng.
Any hint to solve this bug?

Regards,
Zied
![screenshot](https://user-images.githubusercontent.com/4332251/36676201-47b1a01a-1b0b-11e8-8757-361f1c0741ac.png)

aouinizied:
here the log when nprobe is configured with active timeout of 3 secs:

26/Feb/2018 15:44:40 [CollectorInterface.cpp:220] {"8":"192.168.1.103","12":"84.39.50.197","10":65535,"14":65535,"2":150361,"1":225240778,"22":1519656276,"21":1519656278,"7":34815,"11":5001,"6":0,"4":17,"5":0,"16":0,"17":60940, "57943":"192.168.1.103"} [msg_id=24]
26/Feb/2018 15:44:41 [CollectorInterface.cpp:220] {"8":"192.168.1.103","12":"84.39.50.197","10":65535,"14":65535,"2":79,"1":118342,"22":1519656278,"21":1519656278,"7":34815,"11":5001,"6":0,"4":17,"5":0,"16":0,"17":60940, "57943":"192.168.1.103"} [msg_id=27]
26/Feb/2018 15:44:43 [CollectorInterface.cpp:220] {"8":"192.168.1.103","12":"84.39.50.197","10":65535,"14":65535,"2":147754,"1":221335492,"22":1519656279,"21":1519656281,"7":60820,"11":5001,"6":0,"4":17,"5":0,"16":0,"17":60940, "57943":"192.168.1.103"} [msg_id=32]
26/Feb/2018 15:44:45 [CollectorInterface.cpp:220] {"8":"192.168.1.103","12":"84.39.50.197","10":65535,"14":65535,"2":206475,"1":309299550,"22":1519656282,"21":1519656284,"7":60820,"11":5001,"6":0,"4":17,"5":0,"16":0,"17":60940, "57943":"192.168.1.103"} [msg_id=38]
[...]

PR #1695, opened 2018-03-09
Fix issue #1675

aouinizied:
Changing the minimum value from 5 secs to 1 sec fixes the issue #1675.


simonemainardi:
Thanks for providing this contribution.

I wanted to highlight that you can control the stats update frequency directly from the preferences as:
image


By changing the preference above, you can control the `Flow::update_hosts_stats` execution frequency. Commit a5952670daea6ad8ef8ae8d4efde276ee330b0b0 removes the minimum hard-coded value of 5 seconds for ZMQ interfaces, so you should be OK now. If you need a very small refresh period, just set 1 sec in the preference. For this reason, I am closing this PR. Thanks again!

PR #1726, opened 2018-04-01
Support remote fritzbox access in fritzdump script

pschaub:
This will add support for remote capturing of your Fritz!Box traffic in `fritzdump.sh`.

Currently the script only supports local Fritz!Box devices.
This PR will add support to use remotely located Fritz!Box devices.

Changes:
1. Ignore SSL certificate warnings on `curl` command. Needed because the Fritz!Box creates an own self-signed SSL certificate.
2. Ignore SSL certificate warnings on `wget` command. Needed because the Fritz!Box creates an own self-signed SSL certificate.
3. Support changing the protocol of the fritzbox url from `http` to `https`.

### DIFF
--- a/tools/fritzdump.sh
+++ b/tools/fritzdump.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # The is the address of the router
-FRITZIP=192.168.2.1
+FRITZIP=http://192.168.2.1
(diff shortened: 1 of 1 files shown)


cardigliano:
Thank you

PR #1943, opened 2018-08-25
This line causes the build to fail on FreeBSD, so extend the conditio…

madpilot78:
…nal.


simonemainardi:
thank you.

simonemainardi:
cherry-picked to stable as well.

PR #1947, opened 2018-08-27
feat(fritzdump.sh): add sanity checks, user feedback, username argument

bentolor:
This patch changes the CLI interface from `fritzdump.sh <pass>` to `fritzdump.sh <user> <pass>`

It also changes a few settings to more sane defaults, which should enable that the script can run out of the box in many cases

1. the router IP to the router is now resolved via http://fritz.box
2. the default monitored interface is changed to to the WWAN interface 2-0

### DIFF
--- a/tools/fritzdump.sh
+++ b/tools/fritzdump.sh
@@ -1,17 +1,23 @@
 #!/bin/bash
 
-# The is the address of the router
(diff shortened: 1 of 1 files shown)


simonemainardi:
thank you!

PR #2088, opened 2018-10-09
Fixed default path to ndpi include files.

hlein:
nDPI 2.4ish installs headers in ${includedir}/ndpi/, not
${includedir}/ndpi/libndpi. And, I just issued PRs which fix nDPI's
libndpi.pc file. So, ntopng can just believe the -I it gets from
pkg-config.


simonemainardi:
thank you

PR #2101, opened 2018-10-12
Fix race condition while building exposed by removing '.NOTPARALLEL:'.

madpilot78:
In 0b5bbb8 the .NOTPARALLEL directive was removed.

$(TARGET) requires third party library to be compiled before it can start compiling.

Otherwise compilation of ntopng may file because it tries to link to LUAjit before the library has been generated.


simonemainardi:
thank you

Issue #1599, opened 2017-12-11
Host groups are partial (when the number of hosts is high)


emanuele-f:
Hosts groups (e.g. Networks, ASn, Countries) are generated by creating dynamic support arrays to perform the aggregations and sort. For example, NetworkInterface::sortHosts support array has a limit of CONST_MAX_NUM_HITS (32768) entries. When the actual number of hosts is greater that 32768, the aggragated view will only reflect a portion of the actual hosts.

The solution to this is not to increase that limit but rather to rework aggregations to avoid using such big support arrays.

simonemainardi:
The only thing left to remove is 

```
$ egrep -r get_grouped_hosts_data scripts/
scripts//lua/os_stats.lua:print [[/lua/get_grouped_hosts_data.lua?grouped_by=os]]
```

Operating Systems must be handled as another per-interface hash table, e.g., see `include/AutonomousSystemHash.h include/VlanHash.h`

Then, a `get_countries_data.lua` must be implemented and called from the datatable shown in http://develv5:3000/lua/os_stats.lua

For reference, see the implementation of how autonomous systems, or vlans, are handled.

simonemainardi:
Remember to cleanup the code once done. That is, throw get_grouped_hosts_data.lua and related functions which will be no longer used.

simonemainardi:
You should implement a method like `as = iface->getAS(&ip, true /* Create if missing */, true /* Inline call */))` for the operating system.

Each instance of `Host` will have a pointer to the OperatingSystem, .i.e., OperatingSystem os will become a pointer to an entry in the operating systems hash. Set this pointer in `setOS` and don't forget to increase/decrease reference counters.

PR #2104, opened 2018-10-15
Refactor NetworkInterface and Grouper code to remove the limit at CON…

madpilot78:
…ST_MAX_NUM_HITS for Hosts and Hosts and HostsGroup.

This should address issue #1599.


madpilot78:
As a sidenote this same fix could be applied to the following methods too:

NetworkInterface::sortMacs
NetworkInterface::sortASes
NetworkInterface::sortCountries
NetworkInterface::sortVLANs

lucaderi:
@simonemainardi @emanuele-f Any news?

include/Grouper.h
```diff
@@ -40,26 +40,30 @@ struct groupStats{
   char country[3];
 };
 
-class Grouper {
- private:
-  sortField sorter;
-
-  int table_index;
+struct group {
   int64_t group_id_i;
```
simonemainardi: these group_id can be put together in a union as they're never used together. The only values set depends on the sorter.

src/Grouper.cpp
```diff
@@ -25,191 +25,283 @@
...
+/**
+ * Returns group index. Calls newGroup() if no matching group is found.
+ *
+ * Returns -1 for unsupported sorting criteria, causes host to be skipped.
+ */
+int32_t Grouper::inGroup(Host *h) {
+  if(h == NULL)
+    return -1;
 
+  for(int32_t i = 0; i < numGroups; i++) {
```
simonemainardi: this `for` raises the complexity of the code of a squared factor. Let's say we have N hosts and N groups (worst-case). Function `inGroup` is called for every host and, for every hosts, all the groups are visited with the `for`. So, in the worst-case, the complexity goes from O(N) to O(N^2). It's too much to pay this price.

Note the original idea of class `Grouper`: sort hosts one time according to the group key: then you'll be able to visit them and detect groups automatically simply by checking a change in the group key.

src/Grouper.cpp
```diff
@@ -25,191 +25,283 @@
...
+  if(newg == NULL) {
+    ntop->getTrace()->traceEvent(TRACE_WARNING, "Not enough memory");
+    return -1;
   }
+  groups = newg;
 
-  if(group_label){
-    free(group_label);
-    group_label = NULL;
+  groups[numGroups] = (group *)malloc(sizeof(struct group));
```
simonemainardi: could have used calloc

src/Grouper.cpp
```diff
@@ -25,191 +25,283 @@
...
   char buf[32];
+  group **newg;
 
   if(h == NULL)
     return -1;
 
-  if(group_id_s){
-    free(group_id_s);
-    group_id_s = NULL;
+  newg = (group **)realloc(groups, sizeof(struct group) * (numGroups + 1));
```
simonemainardi: doing a realloc every time a new group is created could be an expensive operation

simonemainardi (commented):
Thanks for the contribution. However, the complexity of the `Grouper` implementation proposed is, without considering the cost of the sort, O(N^2) rather than O(N), with N the number of hosts. Please, review the comments and make sure to keep the complexity O(N).

include/Grouper.h
madpilot78: Good suggestion.

src/Grouper.cpp
madpilot78: Will require some thought on how to avoid the complexity. (problem is related to the following two points)

src/Grouper.cpp
madpilot78: Previous code iterated on the already sorted (but partial, for big sets) list, while my code grabs objects from the hash table as they come, unsorted. So my code has no way of knowing in advance how much memory to allocate for them.

I could make it reallocate in bunches of objects(ten at a time for example) instead of reallocating one by one.

Since the system malloc and friends usually already allocate in bunches bigger than the actually requested memory I thought leaving such preallocation work to them was more reasonable.

src/Grouper.cpp
madpilot78: BTW, while the added complexity is significant and I accept that it needs to be fixed, it should be kept into account that these changes remove the complexity related to the sorting of the whole set that used to happen in NetworkInterface::getActiveHostsGroup before working on the resulting set in the Grouper. My code only performs one pass on the whole set.

I'm evaluating methods to keep the group list in memory, so as to simplify the search.

madpilot78:
I have merged the string and integer IDs in a union.

I also have reworked the Grouper to keep the list of groups sorted and use a binary search to find existing groups.

src/Grouper.cpp
simonemainardi: sorting has a logarithmic complexity, and in the worst-case, the whole set or the groups have an equal number of members so the complexity would still be the same. So you don't have to worry too much if you have to do just 1 sort.

src/Grouper.cpp
simonemainardi: I am not sure why you didn't consider the use of a map (that is, an hash stable) to do this kind of activity.

madpilot78: I did consider it but preferred at least to try keeping a "flat" data model before going for that.

That;s the next step in my strategy.

madpilot78:
Something went wrong with the rebase. I'll be coming back once I fix that.

lucaderi:
This patch is pending for a long time and it is now conflicting with the current code tree. Needs to be reworked and comments addressed

madpilot78:
I did address the comments in the latest patch.

I need a little time to rebase the patch on the dev branch. I'll reopen the bug once done.

Issue #1863, opened 2018-07-18
Add Radius Support


lucaderi:
Add Radius support similar to how LDAP is working today

simonemainardi:
please review https://github.com/ntop/ntopng/pull/2177

emanuele-f:
Integrated, now `radcli` needs to be added to the packages dependencies before closing this issue.

emanuele-f:
Dependencies added in 20e2542b4b0dda33bb005d002dc58a0724018635 .

PR #2169, opened 2018-11-20
Add radius support

jerome-42:
Hi,

I made this patch as I need it (related #1863)
It works, but I can not set up the GUI, as the compiled community source code doesn't include Authentication tab on /lua/admin/prefs.lua?tab=auth

I try to keep the same coding conventions as yours.

Also, I will be glad if ldap or radius can set CONST_STR_USER_NETS (ntopng.user.%s.allowed_nets) as local users can.

Feel free to include it to community or pro or entreprise edition.

Unfortunately radcli does not expose rc_read_dictionary_from_buffer, with it we can embedded the dictionnary file into source code and not rely on external file.

Best Regards,


jerome-42:
Maybe a http authenticator is needed to ?

simonemainardi:
now part of https://github.com/ntop/ntopng/pull/2177

PR #2174, opened 2018-11-21
Http authenticator

jerome-42:
Hi,

This is a http authenticator.

NTOP will request a POST to an url with a JSON content:
{ "user": "username", "password": "password" }

User is granted if webservice return code 200, anything else
user is not granted. Content-type and answser content are not checked.

If you want to grant a user as an administrator, webservice must
reply a JSON like:
{ "admin": true }

Sorry, I miss to change the branch when I wrote the radius authenticator :(

I hope to add support of ntopng.user.%s.allowed_nets throught this authenticator, maybe a second revision of this authenticator :)

Jérôme


simonemainardi:
Hi, thanks for taking time to contribute to this project. We really appreciate.

I've noticed that this pull request has a great overlap with your other pull request: https://github.com/ntop/ntopng/pull/2169

Do you think you can merge the code and make a single pull request?

simonemainardi:
now part of https://github.com/ntop/ntopng/pull/2177

PR #2196, opened 2018-12-03
Http allowed nets

jerome-42:
In http authenticator, permit injection of allowed nets.

Maybe add a function to free the HTTPAuthenticator content (in Utils.cpp file) ?


emanuele-f:
I think that this is a very specific need. You basically need a way to configure ntopng via an API, but this cannot be limited to the HTTP authenticator.

jerome-42:
I would like to add to radius too.
Also, what do you think about add others configuration keys to radius / http / ldap authenticator ?
Today, only the local one supports all attributes.

(thanks for pulling radius and http authenticator this last week-end :))

emanuele-f:
I see, I confused it with the global local network, instead you refer to the user allowed networks. This is a list of possible user fields to expose:
- admin/non admin
- allowed interface
- allowed networks
- language

Regarding the Filter-Id on Radius to identify admin (already integrated), is this semantically correct or just an hack to make this work? How would you expose the fields above from radius?

jerome-42:
Radius is a problem. I have duplicated the behavior used in Observium.
The reference [RFC2865](https://tools.ietf.org/html/rfc2865#section-5.11) write:
      This Attribute indicates the name of the filter list for this
      user.  Zero or more Filter-Id attributes MAY be sent in an
      Access-Accept packet.

I have not see other attribute which stick to NTOP definition (admin/simple user). I think Filter-Id is a good attribute.

Allowed interface, allowed networks and language have the same problem, none match a radius attribute.
Then:

- hijack a single text attribute (like Filter-Rule [RFC4849](https://tools.ietf.org/html/rfc4849), concatenate values and parse;
- hijack 3 more text attributes;
- create 4 specific [VSA](https://en.wikipedia.org/wiki/RADIUS#Vendor-specific_attributes) dedicated to NTOP. This is the elegant answer of the problem, but radius server must be setuped in order to understand these new VSA.

emanuele-f:
Thank you for the detailed answer. @simonemainardi what do you think about this?

jerome-42:
Hi,

What do you think about this patch ?

Allowed nets can be binded only with local user or webservice one. Ldap and radius authenticator can be putted aside as unsupported. If users want to set allowed nets, they must implement webservice authenticator.

What do you think about this ?

simonemainardi:
looks good to me. thank you.

simonemainardi:
please, fix the conflicts and resubmit. Sorry for the late reply I missed your earlier comment.

simonemainardi:
@jerome-42 are you able to fix the conflicts so we can include the PR?

jerome-42:
Ohhh sorry, I will fix this in a few days

jerome-42:
@simonemainardi how can I see the conflicts ? I have a small practice on Github. Can you pinpoint me an URL ?

emanuele-f:
These are the steps I would take:

1) Backup your current branch just in case something goes wrong: `git checkout -b backup_branch`
2) Go back to your original branch: `git checkout http-allowedNets`
3) Fetch the new changes: `git fetch origin`, where "origin" is the remote name of the official ntopng
4) Update your branch and solve the conflicts: `git rebase origin/dev`. Follow the instructions on command line to fix them. You can run a `git rebase --abort` if something goes wrong in the meantime.
5) After all the conflicts are resolved (verify with `git status`), push the changes to github (you will have to use `git push --force` since the rebase took place, please double check before doing this).

jerome-42:
Thanks for your help

jerome-42:
Hi @simonemainardi, do you think you can merge this ?
Do not hesitate to tell me if something is wrong.

simonemainardi:
thank you

PR #2313, opened 2019-01-15
X.509 Client Authentication fix, cleanup and doc.

raphpen:
Fixed client cert verification function, must never fail cause, as
stated in SSL_CTX_set_verify(3), TLS/SSL handshake is immediately terminated
if default verification callback function fails, preventing any further login
request, when the client uses an invalid certificate.

Setted a more explicit SSL session context id.

Added documentation at the end of doc/README.SSL.


simonemainardi:
thank you

PR #2326, opened 2019-01-19
Implements Mac stats/data reset and reset at midnight

emanuele-f:
A preference has been added to reset the stats at midnight. Right now, only Mac/Host stats are reset (e.g. not the VLAN/AS stats).


simonemainardi:
thank you

Issue #2435, opened 2019-03-03
Add `libjson` dependency


junquera:
Hi! 

I was updating my docker repository [ntopng-docker](https://github.com/junquera/ntopng-docker) with ntopng 3.8 version, and there were multiple errors related with `json_config.h` when compiling:

```
Makefile:157: recipe for target 'src/VirtualHostHash.o' failed
make[1]: *** [src/VirtualHostHash.o] Error 1
In file included from /ntop/ntopng/third-party/json-c/json_object.h:16:0,
                 from /ntop/ntopng/third-party/json-c/linkhash.h:16,
                 from /ntop/ntopng/third-party/json-c/json.h:22,
                 from /ntop/ntopng/include/ntop_includes.h:126,
                 from src/ThreadPool.cpp:22:
/ntop/ntopng/third-party/json-c/json_inttypes.h:5:10: fatal error: json_config.h: No such file or directory
 #include "json_config.h"
          ^~~~~~~~~~~~~~~
compilation terminated.
```

Adding `libjson-c-dev` fixed this errors, should it be updated in [README.compilation](/ntop/ntopng/blob/dev/doc/README.compilation)?

simonemainardi:
Yes please, feel free to update the readme and send us a pull request for inclusion. Thank you.

PR #2437, opened 2019-03-04
Update README.compilation

junquera:
Added `libjson-c-dev` ubuntu dependency (issue #2435) I've just added in Ubuntu because is the only distribution I have tested, and I'm not sure if is the same name in other distros


simonemainardi:
thank you

PR #2446, opened 2019-03-06
ARP Data Structures

FrancescoStaccini:


simonemainardi:
@FrancescoStaccini do you think we can re-use the already-existing classes `MacHash` and `Mac` to keep the same information and thus keep the matrix information? Did you consider this? That would greatly reduce the size of the commit.

PR #2466, opened 2019-03-20
Added ICMP "port unreachable" counters

SalvatoreCostantino:


src/Flow.cpp
```diff
@@ -88,6 +88,22 @@ Flow::Flow(NetworkInterface *_iface,
       that complains, it means that the srv_host has made a connection that triggered
       the issue and thus it must be accounted in reverse
      */
+    u_int8_t type=icmp_info->getType();
```
simonemainardi, line 91: please abstract logic behind port unreachable in an ICMPinfo class method. E.g, isPortUnreachable().

src/Flow.cpp
```diff
@@ -88,6 +88,22 @@ Flow::Flow(NetworkInterface *_iface,
...
+    u_int8_t code=icmp_info->getCode();
+
+    if(protocol==IPPROTO_ICMP){
+      if(type==3 && code==3){
```
simonemainardi, line 95: Do not hard-code constants. There are macros available for ICMP types and codes. Use them. Look in the ICMPinfo implementation for examples.

include/TimeseriesStats.h
```diff
@@ -30,6 +30,7 @@ class TimeseriesStats: public GenericTrafficElement {
   Host *host;
   u_int32_t total_alerts;
   u_int32_t unreachable_flows_as_client, unreachable_flows_as_server;
+  u_int32_t port_unreachable_flows_as_client, port_unreachable_flows_as_server;
```
simonemainardi, line 33: The point is that actually unreachable_flows_as_client and unreachable_flows_as_server were already containing the port-unreachable-flows of this pull request. 

I would leave  u_int32_t unreachable_flows_as_client, unreachable_flows_as_server.

simonemainardi:
closing as this is a duplication of already-implemented code:

```
  if(icmp_info && icmp_info->getUnreach()) {
    /*
      This is an ICMP flow which contains unreachable information. As is the cli_host
      that complains, it means that the srv_host has made a connection that triggered
      the issue and thus it must be accounted in reverse
     */
    if(cli_host) cli_host->incNumUnreachableFlows(true  /* as server */);
    if(srv_host) srv_host->incNumUnreachableFlows(false /* as client */);
  }
```

If you need something extra on the lua, create another pull request.

PR #2481, opened 2019-03-25
Added new time series

SalvatoreCostantino:


simonemainardi:
thank you

scripts/lua/modules/ts_5min_dump_utils.lua
```diff
@@ -187,7 +187,48 @@ function ts_dump.host_update_stats_rrds(when, hostname, host, ifstats, verbose)
   ts_utils.append("host:unreachable_flows", {ifid = ifstats.id, host = hostname,
 					   flows_as_client = host["unreachable_flows.as_client"],
 					   flows_as_server = host["unreachable_flows.as_server"]},
-		  when, verbose)
+      when, verbose)
+  
+  interface.select(tostring(ifstats.id))
```
simonemainardi, line 192: why the selection again?

scripts/lua/modules/ts_5min_dump_utils.lua
```diff
@@ -187,7 +187,48 @@ function ts_dump.host_update_stats_rrds(when, hostname, host, ifstats, verbose)
...
+   
+  local host_ = interface.getHostInfo(hostname)
```
simonemainardi, line 194: why are you calling `getHostInfo` again? this can be detrimental for the performance. Host info is already passed to this function as a parameter named `host`:

```
function ts_dump.host_update_stats_rrds(when, hostname, host, ifstats, verbose)
```

Calling it again will cause unnecessary extra work and will put extra pressure on the CPU.

scripts/lua/modules/ts_5min_dump_utils.lua
```diff
@@ -187,7 +187,48 @@ function ts_dump.host_update_stats_rrds(when, hostname, host, ifstats, verbose)
...
+              replies_ok_packets = host_["dns"]["rcvd"]["num_replies_ok"],
+              replies_error_packets = host_["dns"]["rcvd"]["num_replies_error"]},
+        when, verbose)
+         
+    ts_utils.append("host:udp_pkts", {ifid = ifstats.id, host = hostname,
+              packets_sent = host_["udp.packets.sent"],
+              packets_rcvd = host_["udp.packets.rcvd"]},
+        when, verbose)
+      
+    local stats = interface.getMacInfo(host_["mac"])
```
simonemainardi, line 221: why are you doing `getMacInfo`? mac addresses are already iterated when calling `callback_utils.foreachDevice` in this very same file....

SalvatoreCostantino: because i can't get arp stats from the table host_

scripts/lua/modules/ts_5min_dump_utils.lua
SalvatoreCostantino: The content of "host_" is different from the content of "host". In "host" i haven't a lot of informations that i've  in "host_"

scripts/lua/modules/ts_5min_dump_utils.lua
simonemainardi: you can:
- decide to store arp counters in ip addresses (ip addresses in arp packets are already dissected) - in which case you'll have such counters in the host.
- leave arp counters in the mac addresses and use host mac address at query time to retrieve mac arp counters

calling this `getMacInfo` causes unnecessary extra work to be done...

scripts/lua/modules/ts_5min_dump_utils.lua
SalvatoreCostantino: I think for a correct call to interface.getHostInfo(). in other scripts the are called together. isn't it necessary?

scripts/lua/modules/ts_5min_dump_utils.lua
simonemainardi: either add what you need in `host` from C, or do these extra calculations using the custom timeseries framework provided (see the docs) so that you'll get the metrics of interest without affecting the ntopng performance for all the users.

scripts/lua/modules/ts_5min_dump_utils.lua
simonemainardi: interface is already selected

scripts/lua/modules/ts_5min_dump_utils.lua
SalvatoreCostantino: Could you explain better the first point, please?

scripts/lua/modules/ts_5min_dump_utils.lua
SalvatoreCostantino: ok, thank you

scripts/lua/modules/ts_5min_dump_utils.lua
SalvatoreCostantino: ok, thank you

scripts/lua/modules/ts_5min_dump_utils.lua
simonemainardi: We already dissect ARP packets to get sender and target destination addresses. Those guys are IPv4s so you can decide to count ARPs for IPs.

simonemainardi:
I decided to revert the commit for the reasons described in the comments: https://github.com/ntop/ntopng/commit/41383aa51b9248dda38ceb1aec3b044dc7cca3a4

Please, do the necessary changes and resubmit. Thanks for the contribution.

PR #2500, opened 2019-04-02
Add Dockerfile based on 18.04 LTS

aaronjwood:
I wanted a newer environment to run ntopng in since [this](https://github.com/lucaderi/ntopng-docker) was last updated 3 years ago.

I'm running this successfully on my server via systemd:
```
[Unit]
Description=ntopng
Requires=docker.service
After=docker.service

[Service]
Restart=always
ExecStart=/usr/bin/docker run --name ntopng --net=host -t -v /ssd/ntopng:/var/lib/ntopng ntopng -l 1 -w 192.168.1.1:5000 -i enp2s0 -i enp3s0 -i enp5s0
ExecStop=/usr/bin/docker stop ntopng
ExecStopPost=/usr/bin/docker rm ntopng

[Install]
WantedBy=multi-user.target
```

If you think the systemd file is helpful let me know and I can commit a template of it in this MR.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,11 +2,14 @@ FROM ubuntu:18.04
 
 ENV DEBIAN_FRONTEND noninteractive
 
-RUN apt-get update && apt-get -y install wget lsb-release gnupg
+# Pass --build-arg TZ=<YOUR_TZ> when running docker build to override this.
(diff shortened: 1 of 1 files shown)


lucaderi:
Thank you

PR #2505, opened 2019-04-02
Fix port unreachable labels

SalvatoreCostantino:


simonemainardi:
must preserve the word "Total" to avoid confusing the series with those that are "Active"

simonemainardi:
thank you

PR #2506, opened 2019-04-02
Fix active contacts time series

SalvatoreCostantino:


simonemainardi:
methods 

```
  virtual u_int16_t getNumActiveContactsAsClient() const  { return stats->getNumActiveContactsAsClient(); };
   virtual u_int16_t getNumActiveContactsAsServer() const  { return stats->getNumActiveContactsAsServer(); };
```

seem to have been replicated equally both on Host and LocalHost. Please reuse them and not that LocalHost inherits from Host.

simonemainardi:
thank you

PR #2517, opened 2019-04-05
Removed icmp net unreachable counters

SalvatoreCostantino:
**Net unreachable** message is nowadays obsolete


simonemainardi:
Thank you

PR #2524, opened 2019-04-10
ARP map

FrancescoStaccini:
graph to visualize the ARP matrix


scripts/lua/arp_matrix_graph.lua
```diff
@@ -0,0 +1,188 @@
...
+dirs = ntop.getDirs()
+package.path = dirs.installdir .. "/scripts/lua/modules/?.lua;" .. package.path
+-- io.write ("Session:".._SESSION["session"].."\n")
+require "lua_utils"
+
+local page_utils = require("page_utils")
+
+interface.select(ifname)
+
+if(ntop.isnEdge()) then
```
simonemainardi, line 14: Why is this needed? Why you need to touch nEdge?

scripts/lua/arp_matrix_graph.lua
```diff
@@ -0,0 +1,188 @@
...
+  package.path = dirs.installdir .. "/pro/scripts/lua/nedge/modules/?.lua;" .. package.path
+  local nf_config = require("nf_config"):readable()
+
+  if nf_config.isFirstStart() then
+    print(ntop.httpRedirect(ntop.getHttpPrefix().."lua/pro/nedge/system_setup/interfaces.lua"))
+    return
+  end
+end
+
+if(ntop.isPro()) then
```
simonemainardi, line 24: Why you need this if? Which kind of checks are you doing?

scripts/lua/arp_matrix_graph.lua
```diff
@@ -0,0 +1,188 @@
...
+
+page_utils.print_header()
+
+-- NOTE: in the home page, footer.lua checks the ntopng version
+-- so in case we change it, footer.lua must also be updated
+active_page = "home"
+dofile(dirs.installdir .. "/scripts/lua/inc/menu.lua")
+
+
+ifstats = interface.getStats()
```
simonemainardi, line 45: missing `local`s. Must declare variables as local

scripts/lua/modules/timeseries/custom/ts_minute_custom.lua
```diff
@@ -0,0 +1,125 @@
+--
```
simonemainardi, line 1: - Can't commit file ts_minute_custom.lua. This file name left for the users who decide to use custom timeseries files. In case, you should create (and commit) another file, e.g., `ts_arp_matrix.lua`
- Why you need this file? Is seems basically a test. Do you just need to add arp stats? In case re-use the existing file when we add stats on the macs.

FrancescoStaccini:
Thanks Simone for the review. I will clean and fix the code.

PR #2540, opened 2019-04-16
ARP Map

FrancescoStaccini:


simonemainardi:
Assuming you haven't touched the chart library`d3.v4`, can you use the minified version `https://d3js.org/d3.v4.min.js` which is smaller, resulting in faster load times?

FrancescoStaccini:
@simonemainardi Let me check

scripts/lua/host_details.lua
```diff
@@ -687,6 +687,45 @@ end
    <button class="btn btn-default" onclick="$('#reset_host_stats_dialog').modal('show')">]] print(i18n("host_details.reset_host_stats")) print[[</button>
    </td></tr>]]
 
+
+   --==========================WIP=======================================
```
simonemainardi: is this WIP expected? do you still have to check if the arp matrix is enabled? In case it is very easy, you can simply read boolean `is_arp_matrix_generation_enabled` from the prefs in lua.

simonemainardi:
thank you

PR #2599, opened 2019-05-20
Timeseries for local talkers (mac devices)

FrancescoStaccini:
Timeseries for the number of local talkers (as client/server) and the related typologies of devices, divided for categories


include/Mac.h
```diff
@@ -40,6 +40,12 @@ class Mac : public GenericHashEntry {
     char * dhcp; /* Extracted from DHCP dissection */
   } names;
 
+  struct{
+    u_int32_t as_server;
+    u_int32_t as_client;
+  }talkers; /*local talkers*/
+  DeviceTypeCounters dev_t_counters = {0,0,0,0,0,0,0,0,0,0,0,0,0};/*type of the devices this mac talked to*/
+
```
simonemainardi, line 48: Initialization inside class declaration. Awkward and also not accepted by all compilers. Use header files only for declarations. Do the necessary initializations inside class members definitions in the corresponding cpp file.

include/Mac.h
```diff
@@ -40,6 +40,12 @@ class Mac : public GenericHashEntry {
     char * dhcp; /* Extracted from DHCP dissection */
   } names;
 
+  struct{
+    u_int32_t as_server;
+    u_int32_t as_client;
+  }talkers; /*local talkers*/
+  DeviceTypeCounters dev_t_counters = {0,0,0,0,0,0,0,0,0,0,0,0,0};/*type of the devices this mac talked to*/
```
simonemainardi, line 47: wasting a huge amount of memory for every Mac... not necessary. You have the matrix, you can keep everything there and obtain the counters simply by navigating  the matrix you already have.

include/Mac.h
```diff
@@ -176,6 +200,10 @@ class Mac : public GenericHashEntry {
     else
       return(false);
   }
+  inline void incTalkersAsClient()  { talkers.as_client++; }
```
simonemainardi, line 203: word Talkers is not self-explanatory per se. Use a name which explains the behavior of these functions.

include/ntop_typedefs.h
```diff
@@ -543,6 +543,22 @@ typedef enum {
...
+  u_int32_t laptop;
+  u_int32_t tablet;
+  u_int32_t phone;
+  u_int32_t tv;
+  u_int32_t networking;
+  u_int32_t wifi;
+  u_int32_t nas;
+  u_int32_t multimedia;
+  u_int32_t iot;
+} DeviceTypeCounters;
```
simonemainardi, line 560: not necessary, see above

src/Mac.cpp
```diff
@@ -228,6 +229,22 @@ void Mac::lua(lua_State* vm, bool show_details, bool asListElement) {
 
   lua_push_uint64_table_entry(vm, "pool", get_host_pool());
 
+  lua_push_uint64_table_entry(vm, "talkers.asClient", getNumTalkerAsClient());
```
simonemainardi, line 232: don't clobber the lua of the Mac with all this stuff. Just push non-zero entries only when necessary, that is, when the arp matrix is enabled.

simonemainardi (commented):
Apart from the other comments, the big issue is that you are wasting a ton of memory for every Mac. You already have the ARP matrix in memory, so you already know the `DeviceTypeCounters` simply by navigating this matrix. There's no need to keep the `DeviceTypeCounters` for every Mac.

FrancescoStaccini:
Ok, thanks for the code review

FrancescoStaccini:
I will open a new pull request as soon as i finished the rework. Thanks again

PR #2651, opened 2019-06-11
Fix inconsistency between total flows counter and icmp flows counters

SalvatoreCostantino:
in my opinion, if a flow is counted as ICMP server flow, it should be counted as server flow (now it is counted as client flow)


simonemainardi:
thank you

PR #2784, opened 2019-08-24
Tirocinio Simone Ricci

SimoneRicci97:
Added metrics calculated by ntop as agreed by mail.


lucaderi:
Changes have been already implementented

PR #2885, opened 2019-09-26
Add extra detection function

emanuele-f:


emanuele-f:
Pull request will be manually merged after fixing some issues.

Issue #2886, opened 2019-09-29
nDPI: no hostname logging of http when dump flows to logstash or elasticsreach


Zeke0n:
as we can see, when running `ndpiReader` is able to captured HTTP hostname, but `HTTP_HOST` of log field is empty even ntopng running under `--dump-hosts=all` mode, still not able to getting `HTTP_HOST`.

how can I get all `HTTP_HOST`?

eg: 
```
	48	TCP 192.168.10.100:53430 -> 202.89.233.101:80 [proto: 7.212/HTTP.Microsoft][cat: Web/5][10 pkts/736 bytes -> 0 pkts/0 bytes][Host: www.bing.com][bytes ratio: 1.000 (Upload)][IAT c2s/s2c min/avg/max/stddev: 0/0 20.7/0.0 94/0 34.6/0.0][Pkt Len c2s/s2c min/avg/max/stddev: 54/0 73.6/0.0 119/0 24.2/0.0][PLAIN TEXT (HEAD / HTTP/1.0)]
	177	TCP 192.168.10.100:65348 -> 180.101.49.11:80 [proto: 7.137/HTTP.GenericProtocol][cat: Web/5][14 pkts/966 bytes -> 0 pkts/0 bytes][Host: www.baidu.com][bytes ratio: 1.000 (Upload)][IAT c2s/s2c min/avg/max/stddev: 0/0 3.5/0.0 24/0 7.6/0.0][Pkt Len c2s/s2c min/avg/max/stddev: 54/0 69.0/0.0 120/0 22.2/0.0][PLAIN TEXT (HEAD / HTTP/1.0)]
```

log sample:
```json
{
    "@timestamp": "2019-09-29T05:10:55.0Z",
    "type": "ntopng",
    "IN_SRC_MAC": "00:00:00:00:00:00",
    "OUT_DST_MAC": "00:00:00:00:00:00",
    "IPV4_SRC_ADDR": "192.168.10.100",
    "IPV4_DST_ADDR": "180.101.49.12",
    "L4_SRC_PORT": 42654,
    "L4_DST_PORT": 80,
    "PROTOCOL": 6,
    "L7_PROTO": 137,
    "L7_PROTO_NAME": "HTTP.GenericProtocol",
    "TCP_FLAGS": 27,
    "IN_PKTS": 12,
    "IN_BYTES": 1044,
    "OUT_PKTS": 0,
    "OUT_BYTES": 0,
    "FIRST_SWITCHED": 1569733852,
[...]

PR #2887, opened 2019-09-29
Allow dump-flows logging HTTP hostname when available

Zeke0n:
I did tested and working will, for issue #2886
QQ20190929-161521@2x


Zeke0n:
fixed in commit: https://github.com/ntop/ntopng/commit/d56ed5870c097038a78e12610a78befc336f0e6c

PR #3156, opened 2019-12-27
Script table

gabryon99:
Add config list and script list for ntopng callbacks.


simonemainardi:
This pull request is not acceptable in this state. There are 1,054 files changes with more than a half million lines of code.

Please, rework the pull request, do a rebase of the most recent dev on top of your branch if necessary, and then resubmit.

The pull request should only contain what you've changed/added and nothing else.

PR #3158, opened 2019-12-28
FritzBox: Add info on usage with password-only auth

lukaschristmann:
The FritzBox interface works with an implicit user called dlsf-config, so it would be easier for users to stay with password-only auth and just enter dslf-config as username.
I updated this in fritzdump.sh and README.fritzbox


simonemainardi:
thank you

Issue #3154, opened 2019-12-27
Community edition dropdowns are poorly styled


simonemainardi:
In community edition, dropdowns look like

![image](https://user-images.githubusercontent.com/3397663/71524187-a121ae80-28cc-11ea-92a4-6be67e896263.png)

it is requested to restyle them to make them look as

![image](https://user-images.githubusercontent.com/3397663/71524246-eb0a9480-28cc-11ea-8010-93543986d5c4.png)

gabryon99:
Issue fixed: https://github.com/ntop/ntopng/pull/3176

PR #3176, opened 2019-12-30
Fix for issue #3154

gabryon99:
I have updated the dropdowns to the latest Bootstrap 4.4 standards.
new-dropdown


gabryon99:
Because the dropdown was too long I gave to it an height so it can be scrollable.

PR #3184, opened 2020-01-02
Add localization for text inside scripts_list and config_list

gabryon99:
As the title says, I added the localization for the last strings that weren't translated yet.


simonemainardi:
thank you

PR #3215, opened 2020-01-10
add label for each hooks into scripts table

gabryon99:
Now the script table has a column with labels for each hooks instead of the keys


emanuele-f:
Which issues fixes/implements this PR?

Issue #3243, opened 2020-01-15
Short message abbreviation is not necessary


emanuele-f:
![2020-01-15_11-59](https://user-images.githubusercontent.com/5488003/72428711-2ab6f480-3786-11ea-85bc-70e7a144af33.png)

Issue #3239, opened 2020-01-14
Unable to disable a script


cardigliano:
When I try to disable the script by unchecking "Enabled", the Apply button is disabled and it is not possible to save the configuration.

<img width="842" alt="Screen Shot 2020-01-14 at 18 51 50" src="https://user-images.githubusercontent.com/7548711/72368757-f64c2580-36fe-11ea-9a94-17a0642150e0.png">

cardigliano:
An (empty) error is also occurring when trying to disable a flow script
![IMAGE 2020-01-15 10:25:48](https://user-images.githubusercontent.com/7548711/72421778-70bd8980-3781-11ea-872d-049fd8aa4094.jpg)

PR #3245, opened 2020-01-15
Fix for issue #3239, #3243

gabryon99:
Now the scripts description that are less than 64 characters don't show the tooltip description with title. Also I fixed the bug for the on/off toggle button inside `edit_configset.lua`. I place the badge class inside the tag `a` for the actions inside the table.


simonemainardi:
thank you

PR #3251, opened 2020-01-17
Checks for not bounded values in templates

gabryon99:
As the title says, now inside the templates there is a check for values that are not valid.


emanuele-f:
Please provide a new clean PR

PR #3269, opened 2020-01-20
elephant flows

gabryon99:


emanuele-f:
With the new changes the apply button does not work

gabryon99:
It seems the problem was with `reloadPageAfterPOST` function. I added a parameter called `force_reload`, so when a user doesn't come from another page there isn't the problem anymore.

PR #3276, opened 2020-01-21
longlived templates for scripts

gabryon99:
I inserted the new `Long Lived` template and improved form validation using the browsers integrations.


emanuele-f:
Issues:
- 60 can be used as the minutes value, but it should not (for consistency). Same issue on the hours.
- Excluded applications and categories: shows "undefined" as placeholder

gabryon99:
Issues fixed, they will be available in the next commit.

Issue #3284, opened 2020-01-22
System and Syslog User Script configurations should not have 'Apply To'


simonemainardi:
Only the default configuration is available and editable without any apply to

![image](https://user-images.githubusercontent.com/3397663/72879936-33677780-3cfe-11ea-97a8-5552816eab62.png)

gabryon99:
opened a pr: https://github.com/ntop/ntopng/pull/3289

PR #3289, opened 2020-01-22
fix for ticket #3284

gabryon99:


emanuele-f:
Buggy, sometimes columns are hidden/shown improperly Peek 2020-01-22 12-40

emanuele-f:
Moreover the columns are hidden only after the page has been loaded

gabryon99:
I fixed the problem, now there isn't a bad refresh.

PR #3317, opened 2020-01-24
fix for filtering criteria

gabryon99:
When a user selects a filter category criteria the `Datatable` searched in all scripts despite the Enabled/Disabled/All flag. Now, this issue is fixed.

I'm going the fix the problems related with hard-coded strings


httpdocs/js/config_callbacks/scripts-list-utils.js
```diff
@@ -1220,31 +1230,44 @@ $(document).ready(function() {
...
+            // list element to append inside the dropdown selector
             const $list_element = $(`<li class='dropdown-item pointer'>${c}</li>`);
+            
+            // when a user click the filter category then the datatable
+            // will be filtered
             $list_element.click(function() {
 
-               if (c == 'All') {
-                  $script_table.column(1).search('').draw();
+               if (c == i18n.all) {
```
simonemainardi: again, you cannot rely in a localized string for this check.

gabryon99: Okok, now I use another approach

gabryon99: Okay, I changed the logic. Now, each category inside `scripts_categories` has two fields:
* `label`: it is the category name (string)
* `disableFilter`: it is a boolean value that when it's true it disable category filtering

PR #3330, opened 2020-01-27
html tags inside script description popover

gabryon99:
fix for ticket #3310. click description to click the link inside


emanuele-f:
Multiple popups are now shown, please fix it.

Peek 2020-01-27 15-58

Also popups and titles are mixed together

PR #3366, opened 2020-01-31
[WIN32] Close the socket properly

gvanem:
In Windows, when you call `close (fd)` on an invalid `fd`, the invalid-parameter handler kicks in. 
A socket-`fd` should off-course be closed using `closesocket (s)` (which is a macro for non-`WIN32` targets).

### DIFF
--- a/src/NetworkDiscovery.cpp
+++ b/src/NetworkDiscovery.cpp
@@ -79,7 +79,7 @@ NetworkDiscovery::NetworkDiscovery(NetworkInterface *_iface) {
 
 NetworkDiscovery::~NetworkDiscovery() {
   if(pd)             pcap_close(pd);
-  if(udp_sock != -1) close(udp_sock);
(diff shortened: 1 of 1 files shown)


simonemainardi:
thanks, please, as `closesocket` is a `Winsock.h`function, please wrap your code in an #ifdef WIN32 and leave the standard `close` otherwise

gvanem:
> please wrap your code in an #ifdef WIN32 

Why? There are other `closesocket()` in the same file. And that expands to `close()` for non-Windows
(via `#include "ntop_defines.h"`).

simonemainardi:
Thank you. by the way, what do you use to build the binary for windows?

gvanem:
> what do you use to build the binary for windows?

A home-brewed GNU-makefile. How do you do it?

simonemainardi:
the procedure is basically manual for Windows as we have to package the professional part as well

PR #3367, opened 2020-01-31
[WIN32] An "x:\path" is valid for Windows

gvanem:
One of my most annoying encounter with NtopNG was messages such as `Unable to locate docs dir` when starting the program with a parameter like `--httpdocs-dir c:\NtopNG\httpdocs`.

So allow a path to have a drive letter and a colon.

### DIFF
--- a/src/Ntop.cpp
+++ b/src/Ntop.cpp
@@ -2075,7 +2075,7 @@ char* Ntop::getValidPath(char *__path) {
 #ifdef WIN32
   const char *install_dir = (const char *)get_install_dir();
 #endif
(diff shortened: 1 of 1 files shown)


simonemainardi:
thanks, can you please use a boolean for the flag rather than int?

gvanem:
> boolean for the flag rather than int?

Done.

*Edit*: I guess you did mean `bool`? `boolean` does work on Windows, but fails on Travis.
Someone fix thus type-mess please.

simonemainardi:
thank you

PR #3368, opened 2020-01-31
[WIN32] Fix pthread_mutex_init() implementation

gvanem:
`pthread_mutex_init()` is prototyped with `void * notused` as the last parameter.
Hence let the implementation have the same. 

(this goes to show hardly anybody cares about NtopNG for Windows).


simonemainardi:
thank you

PR #3380, opened 2020-02-04
Cast argument for pcap_sendpacket()

gvanem:
Fix this C++ error:
```
NetworkDiscovery.cpp(749): 
error C2664: 'int pcap_sendpacket(pcap_t *,const u_char *,int)': cannot convert argument 2 from
'char [1024]' to 'const u_char *'
```

Got this error using MSVC-2019. But clang-cl is equally picky.

(I guess `-DMDNS_MULTICAST_DISCOVERY` is seldom used?)

### DIFF
--- a/src/NetworkDiscovery.cpp
+++ b/src/NetworkDiscovery.cpp
@@ -746,7 +746,7 @@ void NetworkDiscovery::discover(lua_State* vm, u_int timeout) {
     for(i=0; query_list[i] != NULL; i++) {
       u_int16_t len = buildMDNSDiscoveryDatagram(query_list[i], sender_ip, sender_mac, msg, sizeof(msg));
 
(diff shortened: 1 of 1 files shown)


simonemainardi:
thank you

gvanem:
*BTW*.  
For an error in `pcap_send_packet()`, `pcap_geterr()` should be used instead of `errno` and `strerror()`.

simonemainardi:
can you submit a PR for this as well? anyway thanks for your contributions so far

gvanem:
> can you submit a PR for this as well?

There are *so* many illegal use of `errno`  and `strerror()` for Windows and/or libpcap, so the job is not only for this file. So I do not where to start fixing this. Years of Windows neglect.

PR #3385, opened 2020-02-06
implemented leaflet.js maps support

gabryon99:
I removed the Google Maps API from the `hosts_geomap.lua` page. Now we are using `leaflet.js` (thanks @ingria) with `OpenStreetMap` as the provider.

The `hosts_geomap.lua` maintains always the same logic of the previous version but now instead it uses OpenStreetMap and leaflet plugin https://github.com/Leaflet/Leaflet.markercluster to create a markers cluster.


emanuele-f (changes requested):
Currently not working as the "Hosts GeoMap" shows a blank map.

PR #3763, opened 2020-04-14
Fix OpenSSL detection on FreeBSD.

madpilot78:
FreeBSD provides a system OpenSSL library in /usr/lib, but that one cannot be detected by pkg-config. Add a condition to detect that one.

If a FreeBSD system has another ssl library installed which provides a pkg-config file(for example via ports collection) for detection that will be used.


lucaderi:
Thank you

PR #3916, opened 2020-05-09
On FreeBSD -lcrypto is also needed with -ssl.

madpilot78:
On FreeBSD -lcrypto is also needed, and depending on OS version is not always automatically picked up.

Reported by user: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=246289


lucaderi:
Thank you

PR #4020, opened 2020-06-05
modified jp language

Yoshihiro-jp:
add/change with modified correct Japanese words and sentences.


simonemainardi:
thank you for your contribution

PR #4039, opened 2020-06-14
Merge Simplified houskeeping loop to 4.0 (Fixes high CPU usage)

madpilot78:
Hi,

I was reported in https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=246929 about ntopng 4./0 high CPU usage in FreeBSD. This was reported in comment to commit 607ba5226383.

After some digging I found out that merging commit 9fabefe to 4.0 does fix the issue.

So I'm sending this pull request to get it merged there.

Please note the commit to dev also included some code about "cping" variables, which looks like some functionality not present it 4.0, so I adapted the code by removing references to that.

I tested quickly this patch in FreeBSD and it seems to work fine.

I asked @mimugmail to also test it. Waiting for his report.

Thanks!

### DIFF
--- a/include/Utils.h
+++ b/include/Utils.h
@@ -58,6 +58,7 @@ public:
   static u_int32_t hashString(const char * const s);
   static float timeval2ms(struct timeval *tv);
   static float msTimevalDiff(const struct timeval *end, const struct timeval *begin);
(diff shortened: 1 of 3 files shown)


simonemainardi:
thank you

PR #4079, opened 2020-06-24
Added NST contribution scripts for ntopng data extraction...

rwhalb:


scripts/lua/rest/v1/get/contrib/nst/nst_export_flows_data.lua
```diff
@@ -0,0 +1,498 @@
+--
```
simonemainardi, line 1: This file has the same behavior of `scripts/lua/rest/v1/get/flow/active.lua` which should be used instead of this. Minor necessary modifications should go into `scripts/lua/rest/v1/get/flow/active.lua`

scripts/lua/rest/v1/get/contrib/nst/nst_export_ip_data.lua
```diff
@@ -0,0 +1,154 @@
+--
```
simonemainardi, line 1: This file must be merged with `scripts/lua/rest/v1/get/host/data.lua` as it has the same functionality.

No extra field names should be defined, e.g., `tdb`, `tdp`. Everything should come as-is directly from lua to avoid duplication.

In case it is necessary to redefine fields (could be useful to reduce the size of the response), e.g., `packet.sent->tdp`, an extra `map` should be passed in the request.

scripts/lua/rest/v1/get/contrib/nst/nst_export_ip_text.lua
```diff
@@ -0,0 +1,73 @@
+--
```
simonemainardi, line 1: I don't think this endpoint is necessary at all. It has the same functionality of  `scripts/lua/rest/v1/get/host/data.lua` but it doesn't return a mime type.


This should be removed or, if necessary, `scripts/lua/rest/v1/get/host/data.lua` should be extended in place of having an extra endpoint.

scripts/lua/rest/v1/get/contrib/nst/nst_export_ip_xml.lua
```diff
@@ -0,0 +1,526 @@
+--
```
simonemainardi, line 1: Merge with `scripts/lua/rest/v1/get/host/data.lua`.

Add an extra parameter to `scripts/lua/rest/v1/get/host/data.lua` which specifies type `xml` and respond. Also keep in mind all the other comments for nst_export_ip_data.lua.

scripts/lua/rest/v1/get/contrib/nst/nst_export_ndpi_protos.lua
```diff
@@ -0,0 +1,42 @@
+--
```
simonemainardi, line 1: Endpoint is duplicate of `scripts/lua/rest/v1/get/l7/application/consts.lua`. Not necessary to have this.

scripts/lua/rest/v1/get/contrib/nst/nst_info_text.lua
```diff
@@ -0,0 +1,55 @@
+--
```
simonemainardi, line 1: This is endpoint `scripts/lua/rest/v1/get/system/stats.lua`. Use or modify it if necessary.

scripts/lua/rest/v1/get/contrib/nst/nst_network_load.lua
```diff
@@ -0,0 +1,72 @@
+--
```
simonemainardi, line 1: Endpoint is duplicate of `scripts/lua/rest/v1/get/interface/data.lua`. Merge with it.

simonemainardi (changes requested):
Thanks for your contribution. We really appreciate. Some comments are inline with the code.

Here are additional comments and missing things.

- Rest v1 api always uses `rest_utils.rc(rc, res)` to return status codes, errors, and data. Everything should be extended to use this. For the examples, you can see existing endpoints.
- Tests are missing. Look at directory `tests/rest/` and extend it to test every single endpoint in a programmatic and reproducible way. All tests must pass before submitting a pull request.
- Documentation is missing. Endpoints need to be documented in `doc/src/api/rest/rest-api-v1.json`. Then documentation can be built with `make html` inside `doc/src`
- Examples are missing. Endpoint examples need to be added to `doc/src/api/rest/examples_v1.rst`
- JSON must never be generated manually. This is very error prone and it will cause bugs due to unescaped characters. There's library dkjson which you can use (don't now why it's commented in your code)
- Many endpoints can be merged with existing ones (see the comments). Avoid code duplication as this increases complexity and make the maintenance more difficult.
- The rest api must stay agnostic to the 'user', i.e., i.e., no `rest/nst/` tailored for a specific use should exist. So everything must go under `rest/v1/{get, set,}` and no reference to `nst` should exist. We have many many API users and it's not a scalable or practicable approach to tailor API calls to any of them. It's the other way around.

Let me also add @lucaderi to this discussion.

rwhalb:
Refactoring Using recommendations by authors...

PR #4186, opened 2020-07-20
Added rest script for custom host data with docs and test scripts

rwhalb:
This script allows the query for custom host data from an individual host or all hosts on a monitoring interface. Selective host data fields with optional alias names are allowed.


simonemainardi:
thank you

PR #4282, opened 2020-08-09
Openbsd67

adek05:
A handful of updates:
* deal with incompatible interfaces for packet timeval: OpenBSD may be using bpf_timeval, while other operating systems use just timeval
* Makefile and instruction updates.

### DIFF
--- a/configure.seed
+++ b/configure.seed
@@ -156,7 +156,7 @@ PKG_CHECK_MODULES([NDPI], [libndpi >= 2.0], [
             AC_MSG_RESULT(found $NDPI_LIB)
          else
             AC_MSG_RESULT(not found $NDPI_LIB: compiling)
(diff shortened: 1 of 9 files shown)


simonemainardi (commented):
Unfortunately, the code duplicates several methods, both in their fingerprints and in their bodies. Moreover, it does not even build with travis.

Please, remove any duplicate name, and if you need to adjust the timeval type from linux to openbsd, you should to it using a `typedef`, not duplicating the code.

Please, review and resubmit.

lucaderi:
@adek05 Adrian can you please update this PR for merging ? @simonemainardi sent you some comments we would like you o consider and you also need to update the code as there are some conflicts. Please consider submitting a new PR with these changes. Thank you

Issue #4281, opened 2020-08-07
Add REST API call to Update a User


lucaderi:
Currently ti is not possible to update am existing user, such as changing the language or its role.

It is requested to implement a new REST endpoint for editing existing users.

PR #4302, opened 2020-08-14
Add edit user endpoint

mcodato:
This PR refers to ticket #4281 

* Add endpoint to edit a user
* Add some specific return codes

### DIFF
--- a/scripts/lua/rest/v1/edit/ntopng/user.lua
+++ b/scripts/lua/rest/v1/edit/ntopng/user.lua
@@ -1,0 +1,91 @@
+--
+-- (C) 2013-20 - ntop.org
+--
+
+local dirs = ntop.getDirs()
+package.path = dirs.installdir .. "/scripts/lua/modules/?.lua;" .. package.path
+package.path = dirs.installdir .. "/scripts/lua/modules/pools/?.lua;" .. package.path
+
+require "lua_utils"
+local json = require ("dkjson")
+local rest_utils = require("rest_utils")
+
+--
+-- Edit a ntopng user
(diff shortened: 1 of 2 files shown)


scripts/lua/rest/v1/edit/ntopng/user.lua
```diff
@@ -0,0 +1,91 @@
...
+--
+-- NOTE: in case of invalid login, no error is returned but redirected to login
+--
+
+sendHTTPHeader('application/json')
+
+local rc = rest_utils.consts_ok
+local res = {}
+
+if not haveAdminPrivileges() then
```
lucaderi, line 25: Here we also could allow users to modify themselves (i.e. mario can modify mario), but this would complicate things as we need to prevent it from changing admin privileges (such as the interfaceId for instance). For now we'll leave it as is

scripts/lua/rest/v1/edit/ntopng/user.lua
```diff
@@ -0,0 +1,91 @@
...
+local all_users = ntop.getUsers()
+if(all_users[username] == nil) then
+    -- User doesn't exist
+    print(rest_utils.rc(rest_utils.consts_user_does_not_exist, res))
+    return
+end
+
+if(host_role ~= nil) then
+    if(not ntop.changeUserRole(username, host_role)) then
+        print(rest_utils.rc(rest_utils.consts_edit_user_failed, res))
```
lucaderi, line 54: The main problem of all these calls to ntop.changeXXX is that in case of failure we do a partial change without a rollback. Somehow this is acceptable albeit not desirable.

lucaderi (commented):
In this PR the documentation has not been updated. In order to further delay this PR to be merged I will update it, For the future please don't forget to do it

PR #4567, opened 2020-10-11
Applied suggested changes

DanieleZulberti:
Added use of local variables. Used flow.getInfo() and srv.ip instead flow.getServerKey(). Now the for breaks when a match is found.


lucaderi:
Thanks

PR #4711, opened 2020-11-12
Fixes ntop#4709 notify when a new device connects to the network

MatteoBiscosi:


simonemainardi:
this PR is a 99% cut-and-paste of https://github.com/ntop/ntopng/blob/dev/scripts/plugins/alerts/network/device_connection_disconnection/user_scripts/interface/device_connection_disconnection.lua

it is requested to either put functions in common between those two user scripts or slightly extend the `device_connection_disconnection.lua`

MatteoBiscosi:
Pull request closed because this plugin is already implemented in device_connection_disconnection.lua

PR #4726, opened 2020-11-15
fix #4703

### LINKED ISSUE #4703: public scope datasources are not working
https://github.com/ntop/ntopng/blob/dev/src/HTTPserver.cpp#L969-L979
Record the code, a public scope datasource can be accessed by a anonymous user. But in fact it doesn't work.

Because of the datasource URI `/datasources/*ds_hash*` is not a `static resource url`. 
https://github.com/ntop/ntopng/blob/dev/src/HTTPserver.cpp#L1050

RicterZ:
Treat public scope datasources/widgets as authorized request.


lucaderi:
Thank you

PR #4728, opened 2020-11-16
Fixes #4709 unexpected new device plugin

### LINKED ISSUE #4709: Notify when a new device connects to the network
ntopng has already a plugin "Device Connection/Disconnection" for notifying when a device connects/disconnects. What is required is to create a new plugin that triggers an alert when
- a new device connects
- such device has never been observed before

In the plugin page, the user should be able to report a list of (comma separated) mac addresses of already known MACs for which no alert should be triggered.

This ticket replaces #4007

MatteoBiscosi:


simonemainardi (commented):
There's still significant cut-and-paste of existing functions. Need to create a `module/` to reuse common functions.

Do as follow:

- place `unexpected_new_device.lua` under `scripts/plugins/alerts/network/device_connection_disconnection/user_scripts/`
- create a `scripts/plugins/alerts/network/device_connection_disconnection/modules` folder and add a utility file such as `connection_disconnection_utils.lua` and put currently cut-and-pasted functions there (make them local, e.g., connection_disconnection_utils.XXX)
- include the module using `plugins_utils.loadModule`. See for example `local active_monitoring_utils = plugins_utils.loadModule("active_monitoring", "am_utils")`
- refactor `device_connection_disconnection.lua` and `unexpected_new_device.lua` to use functions placed in `connection_disconnection_utils.lua`

simonemainardi:
Review according to suggested changes and update.

Issue #4816, opened 2020-12-08
Assertion in enqueue


vimusov:
ntopng 4.2 fails with assertion in enqueue.

Arguments: `ntopng -s -l 1 -v 6 -i ens4`

Assertion: `/usr/include/c++/10/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = Flow*; _Alloc = std::allocator<Flow*>; std::vector<_Tp, _Alloc>::reference = Flow*&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.`

Backtrace from gdb:
```
Thread 40 "inline ifid 0" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fff5f7fe640 (LWP 16322)]
0x00007ffff732bbc5 in raise () from /lib64/libc.so.6
(gdb) bt
#0  0x00007ffff732bbc5 in raise () from /lib64/libc.so.6
#1  0x00007ffff73148a4 in abort () from /lib64/libc.so.6
#2  0x00005555555bf028 in std::__replacement_assert (__file=<optimized out>, __line=<optimized out>, __function=<optimized out>, __condition=<optimized out>) at /usr/include/c++/10/x86_64-redhat-linux/bits/c++config.h:2560
#3  0x0000555555667cef in std::vector<Flow*, std::allocator<Flow*> >::operator[] (this=<optimized out>, __n=<optimized out>) at /usr/include/c++/10/bits/stl_vector.h:1045
#4  std::vector<Flow*, std::allocator<Flow*> >::operator[] (this=0x555556196ed0, this=0x555556196ed0, __n=<optimized out>) at /usr/include/c++/10/bits/stl_vector.h:1043
#5  SPSCQueue<Flow*>::enqueue (flush=true, item=0x7fffe4001900, this=0x555556196e40) at /usr/src/debug/ntopng-4.2-1.1607058371.utm.utm.x86_64/include/SPSCQueue.h:116
#6  NetworkInterface::hookEnqueue(long, Flow*) [clone .constprop.0] (this=0x555555d87660, f=0x7fffe4001900, t=<optimized out>) at src/NetworkInterface.cpp:637
[...]

PR #4819, opened 2020-12-09
Assertion in queue fixed (issue #4816).

vimusov:
Wrong method has been used for reserving space in a queue vector.

### DIFF
--- a/include/SPSCQueue.h
+++ b/include/SPSCQueue.h
@@ -48,7 +48,7 @@ template <typename T> class SPSCQueue {
    */
   SPSCQueue(u_int32_t size, const char * const _name) {
     queue_size = Utils::pow2(size);
-    queue.reserve(queue_size);
+    queue.resize(queue_size);
     tail = shadow_tail = queue_size-1;
     head = shadow_head = 0;
     num_failed_enqueues = 0;


simonemainardi (commented):
I believe `reserve` was the correct method as there's no need to init elements with their default values. Elements are set during enqueue. Why did you change it to `resize`?

vimusov:
@simonemainardi 

I changed method to `.resize` because according to [C++ reference](https://en.cppreference.com/w/cpp/container/vector/reserve) `.reserve` does not change the size of the vector:

> reserve() does not change the size of the vector.

So after `.reverse()` the vector contains 0 items. And below we can see this code:
```
115:    if (tail != next_head) {
116:      queue[shadow_head] = item;
117:      shadow_head = next_head;
```
In line 116 we try to set item by index `shadow_head`. But according to the [C++ reference](https://en.cppreference.com/w/cpp/container/vector/operator_at) the operator `[]` can not do that when the vector has not been resized or filled with items:

> Unlike std::map::operator[], this operator never inserts a new element into the container. Accessing a nonexistent element through this operator is undefined behavior.

Thus we have an undefined behaviour here. Which results to the assertion described in the issue.

vimusov:
By the way I think it would be much better to use `.push_back()` / `.pop_back()` methods instead of using indexes. I even thought to try to rewrite this entire code with these methods but couldn't understand clearly how the `flush` argument in `enqueue` must work.

simonemainardi:
Thanks for your contribution.

> By the way I think it would be much better to use .push_back() / .pop_back() methods instead of using indexes. I even thought to try to rewrite this entire code with these methods but couldn't understand clearly how the flush argument in enqueue must work.

those methods cannot be posted as the queue is used in a lockless-fashion, so only indices can be used.

vimusov:
> those methods cannot be posted as the queue is used in a lockless-fashion, so only indices can be used.

Now I see, thanks!

Issue #4836, opened 2020-12-14
Compile fails on Raspberry Pi (Debian 10.7) due to missing link libraries


dlk3:
I cloned the current versions of nDPI and ntopng and attempted to compile them on a newly installed Raspberry Pi running the latest "10 (buster)" version of the Raspian distribution (Debian 10.7).  The ntopng compile fails during the final linking step.  I've attached a complete log of the process as the [console.log](https://github.com/ntop/ntopng/files/5691895/console.log) file.  The error indicates that it was unable to find libgpg-error.so.0. 

I added "-lgpg-error" to the g++ command and tested it on the command line and that got me past the error, but exposed a second one.  This second error suggested that libatomic was also missing from the g++ command so I added it as well and then the link command completed successfully.

I then modified Makefile.in as follows and recompiled ntopng from scratch successfully:

```
--- Makefile.in.orig	2020-12-14 14:49:45.264274622 -0500
+++ Makefile.in	2020-12-14 14:41:01.671021439 -0500
@@ -17,6 +17,10 @@
 	GET_UTIL = wget --no-proxy -nc
 endif
 
+###### dlk added
+DLK_LIBS = -latomic -lgpg-error
+######
+
 ######
 NDPI_LIB = @NDPI_LIB@
 NDPI_INC = @NDPI_INC@
@@ -106,7 +110,7 @@
 
 ######
 TARGET = ntopng
-NLIBS = $(NDPI_LIB) $(LIBPCAP) $(LUA_LIB) $(LIBRRDTOOL_LIB) $(ZEROMQ_LIB) $(JSON_LIB) $(SNMP_LIB) @MAXMINDDB_LIB@ $(SODIUM_LIB) @HIREDIS_LIB@ @SQLITE_LIB@ @MYSQL_LIB@ @RADCLI_LIB@ @EXPAT_LIB@ @SSL_LIB@ @LINK_OPTS@ @LDFLAGS@ @PRO_LIBS@ $(ZSTD_LIB) -lm -lpthread
[...]

PR #4838, opened 2020-12-15
Add missing link libraries to Makefile.in

dlk3:
#4836


lucaderi:
thanks for the fix. I need to check if these libs are needed on all platforms

PR #4862, opened 2020-12-25
Update jp.lua

Yoshihiro-jp:
Added missing lines compared with en.lua and improved some Japanese words for easy understadings.


simonemainardi:
thank you

PR #4927, opened 2021-01-18
Update ElasticSearch.cpp

AlexDF2017:
Fix in elasticsearch.cpp to correctly manage ES_BULK_BUFFER_SIZE avoiding truncated bulk inserts.


simonemainardi:
Thanks for the contribution. Please, explain in a code comment the rationale, especially for the meaning of the hardcoded 10000.

AlexDF2017 (commented):
Reviewed the exact buffer allocation to avoid last flow truncation.

simonemainardi:
thank you

PR #5078, opened 2021-02-27
Update de.lua

martinscheu:
Hi Luca
Another bunch of German words.. I added missing entries from the en.lua to the de.lua.
regards, Martin


lucaderi:
Thanks

PR #5096, opened 2021-03-09
Formatting num_uses value

fcarli3:
Formatting numeric values in `num_uses` row using `formatValue` function from `ntop_utils.js`


lucaderi:
Thank you

PR #5097, opened 2021-03-10
Add facility to flows dumping to syslog

tomothy83:
In order to separate flow logs from ntopng internal logs, the configuration option is added.

With no facility specified, it is set to `DAEMON` as before.

```bash
ntopng -F=syslog
```

With facility specified, syslog facility is set for dumped flow logs.
The facility is case-insensitive.

```bash
ntopng -F="syslog;local3"
# or
ntopng -F="syslog;LOCAL3"
```


lucaderi:
Thank you!

tomothy83:
@lucaderi Thank you very much for your quick response.
I'm sorry I just noticed I forgot a condition check for windows.
I submitted #5104, please check it.

PR #5103, opened 2021-03-12
Fix table access in REST call

TimotheusBachinger:
... which resulted in alert conts key not beeing transmitted


TimotheusBachinger:
I've observed this in a 4.3 Enterprise version only, so I assume 4.2-stable is not affected.
There may be other call sites which need to be adopted, didn't check all.

lucaderi:
Thank you

PR #5104, opened 2021-03-13
Add ifndef switch for windows.

tomothy83:
Added `#ifndef WIN32` switch. I forgot this in #5097.
Without this, I guess builds will fail on Windows.


lucaderi:
Thank for the fix. I compile ntop for Win from time to time so this patch was not obvious to me immediately

Issue #433, opened 2016-03-07
MAC address are not sent to ElasticSearch 


C-Duv:
I have enabled dump of flow to ElasticSearch server with the following option:

```
-F "es;flows;ntopng-%Y.%m.%d;http://127.0.0.1:9200/_bulk;"
```

It is working fine except I cannot find the MAC addresses in the ES index.
It would be useful to me so that I can analyze traffic per device over long period of time.
How can I add this field?

simonemainardi:
MAC addresses have been added to ES-exported flows. We decided to not export MAC addresses to MySQL to prevent possible space-related issues.

C-Duv:
Thanks: it works :+1:

tieungao88:
How to update IN_SRC_MAC and OUT_DST_MAC on current version (2.2)?

Thanks

simonemainardi:
@tieungao88 update at least to version 2.4

tieungao88:
Thanks :)

PR #5171, opened 2021-04-13
Add MAC and fingerprints to flows in syslog

tomothy83:
Added some fields to dump flows to syslog.

* Source MAC address
* Destination MAC address
* JA3 client fingerprint in TLS flow
* HASSH client fingerprint in SSH flow

Currently, MAC addresses are exported only to Elasticsearch.
From this comment (https://github.com/ntop/ntopng/issues/433#issuecomment-196435289), this restriction is in purpose of preventing possible space-related issues in MySQL dump.
I think it is not a problem in syslog dump.

-----

Question:

* Is it preferable to include fingerprints in ES dump?
* Is it preferable to include server fingerprints? (I thought client fingerprints are more important to be checked and traced.)


lucaderi:
Thank you

Issue #5399, opened 2021-05-24
Alerts: add heading blocks in flow alerts


lucaderi:
Similar to https://it.wikipedia.org/wiki/Pi-hole#/media/File:Pi-hole_Screenshot.png it is requested to modify the alert page (http://XXXX:3000/lua/alert_stats.lua) and, only in the flows tab, add 4 coloured blocks (not as big as in the picture) between the filters field and the chart, that contain the breakdown of
- total alerts
- top alert type
- top victim (victim host with most alerts, if any)
- top attacker (attacker host with most alerts, if any)

lucaderi:
Implemented as requested

Issue #5350, opened 2021-05-17
Hide encryption key from the user interface


lucaderi:
![Screen Shot 2021-05-17 at 15 12 45](https://user-images.githubusercontent.com/4493366/118546843-cf6d8b80-b758-11eb-9a27-614bcdb988b9.png)

For privacy reasons, it is requested to hide the encryption key and add a button that allows the key to be copied in the clipboard.

MatteoBiscosi:
Verified as fixed

simonemainardi:
- Non-admins must not be able to copy the encryption key - the encryption key must not be rendered at all for non-admins
- This is not a string to be localized https://github.com/ntop/ntopng/commit/151a5d2a18d1ecc8fb7150fad24bb555ecedbdc1#diff-1013beffca0e3b562dbb8d3f329f89aa78b46a546d8ca9beda2f7343ffbbc721R3036
- The hidden key says `ifstats.encryption.public_key` but by definition a `public_key` is public so I am not sure why it needs to be hidden.

MatteoBiscosi:
As requested encryption key is not rendered for non Admin user. Added localization for a string not localized. However the string you mentioned has to be localized, cause it was expressively requested to localize it and the same goes for hiding the key. Please ask to @lucaderi for more info about it

simonemainardi:
I am not sure why an all-stars string needs localization.

By the way, there's a JS exception. I believe this is due to non-ZMQ interfaces.

![image](https://user-images.githubusercontent.com/3397663/120658040-63439500-c485-11eb-9a4d-a691e09b21f9.png)

MatteoBiscosi:
Thanks for reporting this error, I didn't get it before. Fixed with the latest update.

PR #5403, opened 2021-05-24
Hides encrypted key and add copy button(issue #5350). Added heading blocks for alert stats  (issue #5399)

VasilisTako:


lucaderi:
We have reviewed your work: nice start, but before merging we require you to do a few changes
- use jquery to fetch JSON data and select fields
- avoid using variable names in Italian (CSS)
- we use bootstrap 5, whereas you are still using BT4 styles
- the blocks you have created should be designed with  https://getbootstrap.com/docs/5.0/components/card/#card-groups and https://getbootstrap.com/docs/5.0/components/card/#card-styles

Issue #5404, opened 2021-05-25
Download Alert Format is Inconsistent


lucaderi:
Clicking on Delete alert in the page below

<img width="666" alt="Screenshot 2021-05-25 at 08 22 14" src="https://user-images.githubusercontent.com/4493366/119448982-54950980-bd32-11eb-84e1-6216432894e6.png">

the following content is returned

{"rc":0,"recordsTotal":1,"rc_str_hr":"Success","rc_str":"OK","rsp":{"records":[{"ip":{"value":"192.168.2.140","shown_label":"MacBookPro","label":"MacBookPro"},"is_attacker":"","score":{"value":20,"color":"#70db70","label":"20"},"family":"host","script_key":"remote_connection","alert_id":{"label":"Remote Connection","value":12},"count":1,"alert_name":"Remote Connection","vlan_id":0,"tstamp":{"value":1621923573,"highlight":"#70db70","label":"08:19:33"},"duration":27,"is_victim":"","msg":{"value":12,"name":"Remote Connection","description":"Host 192.168.2.140 has 1 active flows using a remote access protocol","configset_ref":" <a href=\"/lua/admin/edit_configset.lua?subdir=host&user_script=remote_connection#all\"><i class=\"fas fa-cog\" title=\"Edit Configuration\"></i></a>"},"severity":{"value":3,"color":"#70db70","label":"alerts_dashboard.notice"}}]},"recordsFiltered":1}

The are are several issues
- the file is in JSON and not in text/CSV format as in nIndex. This needs to be fixed
- it contains several HTML tags that are not supposed to stay in a log file but rather on a GUI. This makes the output unfit for any usage

It is request to export data in CSV format as in nIndex, adding a heading column and formatting the column properly so that they can be processed by an automatic tool

simonemainardi:
Implement as follows:

- Convert method `<class>:format_record` into `<class>:format_json_record` everywhere
- remove any "no_html" reference to the methods above
- Convert method `format_record_commom` into `format_json_record_common`
- Adjusts calls everywhere with the new method names
- Implement methods `<class>:format_txt_record`. Those methods will call `<class>:format_json_record` and then will join the returned table JSON table into a text line using a separator specified in the class, e.g., ```

```
jr = format_json_record(...)
jr["new_filed] = "my new txt field"
res = table.concat({jr["alert_type"], jr["alert_name"], ...}, table.text_separator)
```

- Call `<class>:format_txt_record` when the alerts download is requested

[...]

PR #5433, opened 2021-06-01
issue_5404_download_alert_format-inconsistent

lucapruneti:
Implemented alerts download as specified in issue #5404


scripts/lua/rest/v1/get/network/alert/list.lua
```diff
@@ -40,13 +40,25 @@ interface.select(ifid)
 -- Fetch the results
 local alerts, recordsFiltered = network_alert_store:select_request()
 
-for _key,_value in ipairs(alerts or {}) do
-   local record = network_alert_store:format_record(_value, no_html)
+for _, _value in ipairs(alerts or {}) do
+   local record = {}
+
+   if no_html then
```
simonemainardi: To avoid duplicating this `if no_html then ... else ... end` logic inside any of the alert endpoints available, just rely on the `no_html` parameter already existing. You can do a single `alert_store:format_record(_value, no_html)` and put the `if-then-else` only one time inside that function. Even better, you can do a single wrapper for this logic and also to decide between `vanilla_payload_response` and `extended_answer`. This will save a lot of code lines.

scripts/lua/modules/alert_store/alert_store.lua
```diff
@@ -669,67 +669,173 @@ end
...
+function alert_store:to_csv(documents)
+
+   local separator = "|" -- change this to use another separator
+   local csv = ""
+
+   local csv_header = {} -- contains the column heading names
+   for _, value in pairs(self:get_export_record_names()) do
+      csv_header[#csv_header + 1] = tostring(value)
+   end
+   table.sort(csv_header) -- this assure the same csv columns order at each iteration
```
simonemainardi: to sort, see utility functions `pairsByKeys` and `pairsByValues`.

scripts/lua/modules/alert_store/alert_store.lua
```diff
@@ -669,67 +669,173 @@ end
...
+      end
+   end
+
+   return csv
+end
+
+-- Used to escape "'s by to_csv
+function alert_store:escape_csv(s)
+   if string.find(s, '[,"|]') then
+      s = '"' .. string.gsub(s, '"', '""') .. '"'
```
simonemainardi, line 880: this escape is not necessary for csv, it's the sqlite escape. Also note that this escape function doesn't work.

scripts/lua/modules/alert_store/alert_store.lua
```diff
@@ -669,67 +669,173 @@ end
...
+         csv = csv .. row .. '\n'
+      end
+   end
+
+   return csv
+end
+
+-- Used to escape "'s by to_csv
+function alert_store:escape_csv(s)
+   if string.find(s, '[,"|]') then
```
simonemainardi: Also `\n` need to be escaped. Make `local separator = "|"` a class member and use it here in the escape function and also in the csv generation function above.

scripts/lua/modules/alert_store/alert_store.lua
```diff
@@ -669,67 +669,173 @@ end
...
+   return self:format_record(value, false)
+end
+
+-- define the base record names of the document, both json and csv
+-- add a new record name here if you want to add a new base element
+-- name: the actual record name 
+-- export: use only in csv export, true the record is included in the csv, false otherwise
+local BASE_RNAME = {
+   FAMILY = { name = "family", export = true},
+   ROW_ID = { name = "row_id", export = true},
```
simonemainardi: do not include the rowid in the csv export

scripts/lua/modules/alert_store/flow_alert_store.lua
```diff
@@ -317,10 +317,24 @@ end
 
 -- ##############################################
 
+local RNAME = {
+   ADDITIONAL_ALERTS = { name = "additional_alerts", export = true},
+   ALERT_NAME = { name = "alert_name", export = true},
+   DESCRIPTION = { name = "description", export = true},
```
simonemainardi, line 389: description must be present in the csv of any entity, not only on flows.

simonemainardi (changes requested):
Thanks for this contribution. Please, review the comments. Also make sure all tests complete successfully (run `~/ntopng/tests/rest$ sudo ./run.sh`).

scripts/lua/modules/alert_store/flow_alert_store.lua
lucapruneti: @simonemainardi Could you be more precise about this comment, I don't understand what you mean.
Thanks

simonemainardi: Sure. Sorry for not being accurate enough. The full description of an alert should be exported for all alerts. For example, for host alerts, it is desirable to have exported not only the alert type label but also the full alert description. E.g.,

Not only "TCP SYN Scan" but also "devel is a SYN scan attacker [ZZ > YY SYN sent]"

simonemainardi:
Thank you.

PR #5445, opened 2021-06-04
CSV Export

lucapruneti:
Added FLOW fields in flow alerts
Added Description in all the entities
Improved csv field export in multilevel documents


simonemainardi:
Thank you

PR #5713, opened 2021-07-26
Countries host check

### LINKED ISSUE #5665: Raise an alert when a host contacts an abnormal number of countries/ASN
In every local host keep track of the number of different countries and ASN contacted (if GeoIP is disabled this check won't apply) by incrementing a counter (e.g. based on HLL) when a new flow is created. Implement a threshold in the host that checks when a host contacts a high number of countries/ASN (this can be set by implementing a high threshold e.g. with default value 100), as well when its behaviour changes (for this a DES data structure can be used). This ticket will be implemented with a host (not flow) check.

francescoamodeo:
Me and my colleague @paolo-junior-mollica have implemented a check that counts the number of countries contacted and raises an alert if this number exceeds a threshold. HLL counters and DES structures were used to have both static and dynamic thresholds.
resolves #5665


include/LocalHostStats.h
```diff
@@ -43,7 +43,12 @@ class LocalHostStats: public HostStats {
   struct ndpi_hll hll_contacted_hosts;
   double old_hll_value, new_hll_value, hll_delta_value;
   DESCounter contacted_hosts;
-  
+
+  /* Estimate the number of contacted countries using HyperLogLog */
```
simonemainardi, line 48: Quantify, and report, in terms of Bytes, the impact of all these members for each local hosts.

include/LocalHostStats.h
```diff
@@ -43,7 +43,12 @@ class LocalHostStats: public HostStats {
...
+  struct ndpi_hll hll_countries_contacts;
+  double old_hll_countries_value, new_hll_countries_value, hll_delta_countries_value;
```
simonemainardi: What's the rationale of using double here? Countries in the world are < 255, so why don't you use u_int8?

simonemainardi: I am also not sure that using HLL is a good thing here. Again, countries are < 255, did you estimate the memory used by HLL with reference to a plain bitmap for example?

include/host_alerts/CountriesContactsAlert.h
```diff
@@ -0,0 +1,16 @@
+#include "ntop_includes.h"
+
+class CountriesContactsAlert : public HostAlert {
+private:
+    u_int32_t countries_contacts, countries_contacts_threshold;
```
simonemainardi: set to u_int8

include/host_checks/CountriesContacts.h
```diff
@@ -0,0 +1,24 @@
+#ifndef _COUNTRY_CONTACTS_H_
+#define _COUNTRY_CONTACTS_H_
+
+#include "ntop_includes.h"
+
+class CountriesContacts : public HostCheck {
+protected:
+    u_int32_t countries_contacts_threshold;
```
simonemainardi: set to u_int8

scripts/lua/modules/check_definitions/host/countries_contacts.lua
```diff
@@ -0,0 +1,35 @@
...
+      operator = "gt",
+      threshold = 100,
+   },
+
+   gui = {
+      i18n_title = "alerts_thresholds_config.countries_contacts_title",
+      i18n_description = "alerts_thresholds_config.countries_contacts_description",
+      i18n_field_unit = checks.field_units.contacts,
+      input_builder = "threshold_cross",
+      field_max = 65535,
```
simonemainardi: This is unreasonable. There are << 65k countries in the world. Set to 255.

scripts/lua/modules/check_definitions/host/countries_contacts.lua
```diff
@@ -0,0 +1,35 @@
...
+local checks = require("checks")
+local host_alert_keys = require "host_alert_keys"
+
+-- #################################################################
+
+local countries_contacts = {
+   -- Script category
+   category = checks.check_categories.security,
+
+   default_enabled = true,
```
simonemainardi: leave it disabled by default

src/host_checks/CountriesContacts.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+void CountriesContacts::periodicUpdate(Host *h, HostAlert *engaged_alert) {
+    HostAlert *alert = engaged_alert;
+    u_int32_t contacted_countries = 0;
+
+    if ((contacted_countries = getContactedCountries(h)) >= countries_contacts_threshold) {
+        if (!alert) alert = allocAlert(this, h, CLIENT_FAIR_RISK_PERCENTAGE, contacted_countries, countries_contacts_threshold);
+        if (alert) h->triggerAlert(alert);
+    }
+    h->resetCountriesContacts();
```
simonemainardi: This resets every minute - not 5 minutes. Is this the desired behavior?

src/host_checks/CountriesContacts.cpp
```diff
@@ -0,0 +1,28 @@
+#include "ntop_includes.h"
+#include "host_checks_includes.h"
+
+CountriesContacts::CountriesContacts() : HostCheck(ntopng_edition_community, false /* All interfaces */, true /* Exclude for nEdge */, false /* NOT only for nEdge */) {
+    countries_contacts_threshold = (u_int32_t)100;
```
simonemainardi: default must be (u_intXX)-1 so that no threshold is triggered.

simonemainardi (changes requested):
Please, handle according to the comments.

include/host_alerts/CountriesContactsAlert.h
francescoamodeo: fixed

include/host_checks/CountriesContacts.h
francescoamodeo: fixed

scripts/lua/modules/check_definitions/host/countries_contacts.lua
francescoamodeo: fixed

scripts/lua/modules/check_definitions/host/countries_contacts.lua
francescoamodeo: fixed

include/LocalHostStats.h
francescoamodeo: we have estimated that the size of a hll counter initialized with 8 bits is 256 bytes, so we have decided to decrease it to 5 or 6 bits in order to occupy 32 or 64 bytes.  We believe that the error generated by the lower precision should not be a problem in estimating the number of countries contacted.  By doing so we occupy the same memory of a 256 bit bitmap which should correspond to 32 bytes occupied

src/host_checks/CountriesContacts.cpp
francescoamodeo: It was not the desired behavior, in fact we removed the reset from the `periodicUpdate()` and we inserted it in: 
https://github.com/fram112/ntopng/blob/f1e590a7678679c86a9934ffc58a62a6d5e7a5b1/src/LocalHostStats.cpp#L703
The counter now resets every 5 minutes.

src/host_checks/CountriesContacts.cpp
francescoamodeo: fixed

include/LocalHostStats.h
simonemainardi: Referenced commit with the fix still  has double. What is the rationale of using doubles to keep a number <255?

paolo-junior-mollica: There's no rationale, it was my oversight. I'll fix it right now.

include/LocalHostStats.h
```diff
@@ -46,7 +46,7 @@ class LocalHostStats: public HostStats {
 
   /* Estimate the number of contacted countries using HyperLogLog */
   struct ndpi_hll hll_countries_contacts;
-  double old_hll_countries_value, new_hll_countries_value, hll_delta_countries_value;
+  u_int8_t old_hll_countries_value, new_hll_countries_value, hll_delta_countries_value;
```
paolo-junior-mollica, line 50: @simonemainardi actually we fixed it. Is it possible you were reading line 44 instead of 49? Line 44 isn't part of our code.

simonemainardi:
Please unify and provide a single pull request for:

- https://github.com/ntop/ntopng/pull/5714
- https://github.com/ntop/ntopng/pull/5713

As they both involve the same topic and the same issue. Note, however, that the memory used for the implementation is north of 500 Bytes per local host, without the extra memory allocated by HLL. This is too much so I strongly suggest to find a more memory-savvy implementation.

lucaderi:
Please reduce the memory footprint by decreasing the number of bytes in ndpi_hll_init() to 4 or 5 at most so the memory footprint is significantly reduced.

simonemainardi:
~500 Bytes per local host is the sum of the following `LocalHostStats` class members

image

Note that the sum does NOT include the memory allocated dynamically by HLL.

Although using HLL(5) would equal a plain 256 bits bitmap, HLL is approximated so you would pay the same amount of memory for less accurate results.

Finally, it would be desirable to have all the class members highlighted above wrapped in a dedicated class to
- keep `LocalHostStats` cleaner with just a pointer
- Instantiate the class only when behavioral analysis is enabled to avoid paying all this memory every time
- Unify `::lua` methods that now have a great repetition of identical code between countries, ASNs, and host contacts.

PR #5714, opened 2021-07-26
ASN/Countries host check

tanobar:
Pull request referred to #5665


simonemainardi:
Please unify and provide a single pull request for:

- https://github.com/ntop/ntopng/pull/5714
- https://github.com/ntop/ntopng/pull/5713

As they both involve the same topic and the same issue. Note, however, that the memory used for the implementation is north of 500 Bytes per local host, without the extra memory allocated by HLL. This is too much so I strongly suggest to find a more memory-savvy implementation.

Issue #5716, opened 2021-07-27
Raise an alert when a host contacts an abnormal number of host names


lucaderi:
Same as #5665 but counting domain names contacted by a host. In essence in every localHost instantiate a HLL data structure with a few bytes (4 can be enough) and for every flow do
-  char *name = ndpiFlow ? ndpi_get_flow_name(ndpiFlow) : NULL;
- increment the HLL member used to count the number of contacted host names 
- Raise an alert when a high number of names is contacted every 1 (or 5) minutes.
- A default threshold can be 250 (minute) or 1000 (5 minutes)

D0kken:
My colleague @tanobar and I are working for this Issue. Anyway, is the use of a DES data structure still required ?

MatteoBiscosi:
Implemented as requested

PR #5721, opened 2021-07-28
Domain Names cardinality host check

tanobar:
#5716


include/LocalHostStats.h
```diff
@@ -37,10 +37,11 @@ class LocalHostStats: public HostStats {
   /* Estimate of the number of critical servers used by this host */
   Cardinality num_dns_servers, num_smtp_servers, num_ntp_servers;
 
-  /* Estimate the number of contacted hosts using HyperLogLog */
-  struct ndpi_hll hll_contacted_hosts;
-  double old_hll_value, new_hll_value, hll_delta_value;
-  DESCounter contacted_hosts;
+  /* Estimate the number of contacted hosts,asn and country using HyperLogLog*/
```
simonemainardi, line 42: There's class `Cardinality` (see line 38 above) that estimates the number of DNS, SMTP and NTP servers. That class should be used/extended also to do the estimations on hosts and domain names. Indeed,  `Cardinality` already uses hll in its implementation.

include/host_alerts/DomainNamesConnectionAlert.h
```diff
@@ -0,0 +1,45 @@
...
+
+#ifndef _DOMAIN_NAMES_CONNECTION_ALERT_H_
+#define _DOMAIN_NAMES_CONNECTION_ALERT_H_
+
+#include "ntop_includes.h"
+
+class DomainNamesConnectionAlert : public HostAlert {
+ private:
+  u_int32_t num_domain_names;
+  u_int8_t domain_names_threshold;
```
simonemainardi: 255 as maximum value is too low. Think to a web server but also as a normal host doing normal activities.

include/host_checks/DomainNamesConnection.h
```diff
@@ -0,0 +1,47 @@
...
+ * Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ *
+ */
+
+#ifndef _DOMAIN_NAMES_CONNECTION_H_
+#define _DOMAIN_NAMES_CONNECTION_H_
+
+#include "ntop_includes.h"
+
+class DomainNamesConnection : public HostCheck {
```
simonemainardi: - Name this class `DomainNameContacts` to stay consistent with other classes `DNSServerContacts`, `SMTPServerContacts`, and so on.
- Make `DomainNameContacts` son of `ServerContacts` (possibly rename the latter in `Contacts`).
- Also rename the corresponding enums and typedefs
- Do the very same things described above for the corresponding alert class `DomainNamesConnectionAlert`.

scripts/lua/modules/alert_definitions/host/host_alert_domain_names_connection.lua
```diff
@@ -0,0 +1,64 @@
...
+
+--@brief Prepare an alert table used to generate the alert
+--@param one_param The first alert param
+--@param another_param The second alert param
+--@return A table with the alert built
+function host_alert_domain_names_connection:init(metric, value, operator, threshold)
+--Call the parent constructor
+self.super:init()
+
+self.alert_type_params = alert_creators.createThresholdCross(metric, value, operator, threshold)
```
simonemainardi: This is not needed, as the json is generated from the C++ alert class.

src/Flow.cpp
```diff
@@ -97,28 +97,30 @@ Flow::Flow(NetworkInterface *_iface,
...
   if(cli_host) {
     NetworkStats *network_stats = cli_host->getNetworkStats(cli_host->get_local_network_id());
-
     cli_host->incUses(), cli_host->incNumFlows(last_seen, true);
+
     if(network_stats) network_stats->incNumFlows(last_seen, true);
     cli_ip_addr = cli_host->get_ip();
     cli_host->incCliContactedHosts(_srv_ip);
     cli_host->incCliContactedPorts(_srv_port);
+    if(cli_host->isLocalHost() && domain_name!=NULL)  cli_host->addContactedDomainName(domain_name);
```
simonemainardi: No need to check for local hosts

src/Flow.cpp
```diff
@@ -97,28 +97,30 @@ Flow::Flow(NetworkInterface *_iface,
...
   if(srv_host) {
     NetworkStats *network_stats = srv_host->getNetworkStats(srv_host->get_local_network_id());
-
     srv_host->incUses(), srv_host->incNumFlows(last_seen, false);
     if(network_stats) network_stats->incNumFlows(last_seen, false);
     srv_ip_addr = srv_host->get_ip();
-
     srv_host->incSrvHostContacts(_cli_ip);
     srv_host->incSrvPortsContacts(_cli_port);
+    if(srv_host->isLocalHost() && domain_name!=NULL) srv_host->addContactedDomainName(domain_name);
```
simonemainardi: The check should only be for the client, not for the server

src/LocalHostStats.cpp
```diff
@@ -197,6 +215,25 @@ void LocalHostStats::luaHostBehaviour(lua_State* vm) {
 
 /* *************************************** */
 
+
+
+void LocalHostStats::luaDomainNamesBehaviour(lua_State* vm) {
```
simonemainardi: This `LocalHostStats::luaDomainNamesBehaviour` is an identical cut-and-paste of method above `LocalHostStats::luaHostBehaviour`. I believe this can be solved by using class `Cardinality` as suggested in the other review comment.

simonemainardi (changes requested):
Please, review.

include/LocalHostStats.h
tanobar: It's ok using Cardinality for our hll_contacted_domain_names at line 42, but hll_contacted_hosts was not added by us. We don't know if there is actually a reason because Cardinality was not used so we will only change ours to avoid ruining something.

scripts/lua/modules/alert_definitions/host/host_alert_domain_names_connection.lua
tanobar: Do you mean that we have to delete whole "function host_alert_domain_names_connection:init([params])" or use that function without params and replace line 38 with "self.alert_type_params = {}" in its body?

simonemainardi: just leave the `self.super:init()` in the method. That's enough. You are populating `alert_type_params` in C++

PR #5723, opened 2021-07-29
Domain Names host check

tanobar:
#5716 Some issues forced us to close the previous PR. This is a new one with the requested changes.


simonemainardi (commented):
For me this is good to go. Thanks for the contribution.

However, given that we are in the process of doing the 5.0 stable release, I am not sure if @lucaderi wanted to leave this PR on hold until the 5.0 is out so it can be merged in the dev 5.1.

PR #5735, opened 2021-08-06
Fix some HTTP methods in API document

tomothy83:
I **really** appreciate API v2, especially scripts APIs. Wonderful!

Nevertheless, it seems some of HTTP methods and parameter types in [API v2 document](https://www.ntop.org/guides/ntopng/api/rest/api_v2.html) are not correct.

cf. ) [Swagger document about parameter types](https://swagger.io/docs/specification/2-0/describing-parameters/)


simonemainardi:
Thanks for your contribution

Issue #5906, opened 2021-09-20
Packet size checks


lucaderi:
Similar to other checks, implement a new check for spotting hosts that misbehave. Possible thresholds could be

0-128 byte > 30000 pkt/s
129-256 byte > 10000 pkt/s

or just have a single threshold (10000 pkt/s)

lucaderi:
Marco Tranchida <m.tranchida1@studenti.unipi.it> is working to implement this check

MatteoBiscosi:
Implemented as requested

PR #6074, opened 2021-11-11
Implementazione nuovo check, progetto GR

Nexon13:


lucaderi:
Implements https://github.com/ntop/ntopng/issues/5906

include/ntop_typedefs.h
```diff
@@ -465,7 +465,6 @@ typedef enum {
   flow_alert_ndpi_http_suspicious_content     = 63,
   flow_alert_ndpi_dns_large_packet            = 64,
   flow_alert_ndpi_dns_fragmented              = 65,
-  flow_alert_ndpi_invalid_characters          = 66,
```
simonemainardi: This breaks the integration with nDPI. Don't remove it.

include/HostChecksStatus.h
```diff
@@ -31,12 +31,13 @@ class HostChecksStatus { /* Container to keep per-check status (e.g., traffic de
   u_int64_t ntp_bytes;  /* Holds the NTP bytes and is used to compute the delta of NTP bytes across consecutive check calls */
   u_int64_t p2p_bytes;  /* Holds the P2P bytes and is used to compute the delta of P2P bytes across consecutive check calls */
   u_int64_t dns_bytes;  /* Holds the DNS bytes and is used to compute the delta of DNS bytes across consecutive check calls */
+  u_int64_t my_pkt_counter;
```
simonemainardi: Please, give it a meaningful variable name. This is not perl and my doesn't mean anything. For the rest this is good to go.

simonemainardi (commented):
I have seen that you are dividing and multiplying by 60. What is the rationale of this? Are you trying to capture packets per second?

Please, add comments and make it clear.

In case you are tying to handle packets per second, note that the callback execution is not guaranteed to be executed exactly every minute, so you need to snapshot the time and compute the difference if necessary. The time of the previous call is stored in `HostChecksStatus` whereas you can use `time` to get the current time if necessary.

Nexon13:
PktThreshold.cpp takes the value from the ntopng gui, expressed in pkts/sec. The gui still displays 1 minute granularity, but i was told by Luca Deri that the implementation for 1 sec granularity in the gui was not my concern and that the only thing i had to do was taking the value in pkts/min and then calculate the mean value to use as an estimate of the pkts/sec value. Hope this explains the reason behind dividing and multiplying.

PR #6177, opened 2022-01-04
build: respect environment options more

thesamesam:
Very similar to the PR I've created for nDPI: https://github.com/ntop/nDPI/pull/1392.

- Respect `CXX`, `LDFLAGS` (this is particularly helpful for downstreams)
- Use `$(MAKE)` consistently for (very slightly) better parallel builds/installs
- Start using `configure.ac` instead of `configure.seed` (similar to https://github.com/ntop/nDPI/commit/cf931fda6bfb3925555c7bd11d950a886676bcb3).

In addition to being able to easily use e.g. Clang for testing now, it should be easier to test with different `LDFLAGS` (e.g. different `-fuse-ld=` or `-Wl,--as-needed`).

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -6,7 +6,11 @@ datarootdir = @datarootdir@
 SHELL=/bin/sh
 OS := $(shell uname -s)
 PWD=@PWD@
-GPP=@CXX@ -std=c++0x
+
+CXX=@CXX@
+CXXFLAGS?= -O2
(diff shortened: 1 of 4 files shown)


lucaderi:
Thanks for your contribution

thesamesam:
Thanks a lot @lucaderi, it's a huge help for us downstream!

I'll see if I can find a way to do the versioning stuff which you need in an integrated way.

PR #6185, opened 2022-01-10
Update nindex_export_to_ch.sh

nokoya:
exclude header of csv_file


MatteoBiscosi:
Thanks for your contribution

PR #6283, opened 2022-02-07
Addressres test

giorgiozoppi:
- Made Redis optional for address resolution - if we don't have redis we don't use and cache it. At least  it doesnt crash anymore. It will exit explicitly when ntop checks for redis and it makes more sense.
- Added setter method for the script dir  and data_dir since ntop preferences depends on those otherwise it will not 
perform any tests.
- Added test base NtopTestingBase for all test class. We just need to include that class in each test fixture as member :
prefer composition over heritance princple. So using the constructor does his job.
-  Reorganized the test directory for dividing between unit tests (the new ones) and e2e test (your lua scripts and so on)
-  Added tests for address resolution
The next PR will be the thread pool.


cardigliano:
All good. Thank you!

PR #6288, opened 2022-02-08
build: misc tiny fixes

thesamesam:
- Fail configure if ZMQ can't be found (as already intended)
- Avoid Bashism in configure
- Use $(MAKE) not 'make'


thesamesam:
I think there's some other bashism quirks or some side-effects from how configure.ac.in -> configure.ac is done but I've left that alone for now to avoid any risks. Thanks.

lucaderi:
Thank you

PR #6298, opened 2022-02-09
Threadpooltest: first part

giorgiozoppi:
- clang-format test directory
- added new test for the threadpool using pthread_t

Still missing job submission but it will be in another PR.


giorgiozoppi:
It's a bit messy the threadpool. So better understand it better since the threadactivity code is new for me. 
It does too many things. A thread activity should be just a package task to be executed periodically, immediately with a max execution time (deadline). 
I put this in draft. Meanwhile i add to tests a new class called ScopedThreadPool. The idea here is to simplify and avoid code duplication. 

As user of a threadpool i would expect to do the following thing:
-  create the pool in the current scope
-  submit a ThreadActivity
-  shutdown the pool

And nothing else so when C++ goes out of scope the pool destory itself and i am happy.

This is RAAI,  the current interface instead allows this:

- creation with constructor
-  run method then blocks waiting in a queue
- queue a ThreadActivity
-  shutdown


So i had to execute run from another thread for being able to send jobs because when i call run blocks.

ScopedThreadPool is wrapper to ThreadPool that solves the blocking. So ScopedThreadPool:

- when the constructor gets called it executes the run in another thread so to shutdown
- when the destructors delete the internal ThreadPool and it joins the thread created for executing run.
For the moment i expose the ThreadPool with a get method in order to avoid further changes.
It's ok since it's just for avoid replication during the code of testing.

Thank you!

giorgiozoppi:
- used cmake+ninja for speeding up build: but still a bit of work has to be done.
- added coverage support for the tests
For the moment i will focus in having this PR ready asap and having tests. I will leave ntop20 to the future/after having 
a considerable number of unit tests done, in this way i will understand the code base.

ouliuquan:
I think we should go in two parts:

1. Add minimal changes in thread pool

2. Add the minimum changes of CMake, in the way of compilation, Makefile can still work, which is a way for developers to choose.

After completing these two branches, we can make a further roadmap. We should not introduce CMake in the threadpool branch.

giorgiozoppi:
Good.

ouliuquan:
I just compiled threadpooltest without CMakefile and it compiles fine. I would like to know those two:

1. what is the problem with this branch
2. what are we going to achieve?

I am new here, if I know more that will be helpful. Thank you @giorgiozoppi 
 Here is my branch #https://github.com/ouliuquan/ntopng/tree/threadpooltest_minimal (base on your commit: #https://github.com/ntop/ntopng/commit/7f470d0205216e9a716a0f10e657a17abc35cb5b , before adding CMakefile)

giorgiozoppi:
> I just compiled threadpooltest without CMakefile and it compiles fine. I would like to know those two:
> 
> 1. what is the problem with this branch
> 2. what are we going to achieve?
> 
> I am new here, if I know more that will be helpful. Thank you @giorgiozoppi Here is my branch #https://github.com/ouliuquan/ntopng/tree/threadpooltest_minimal (base on your commit: #[7f470d0](https://github.com/ntop/ntopng/commit/7f470d0205216e9a716a0f10e657a17abc35cb5b) , before adding CMakefile)
 @ouliuquan  What we want is a modern C++, robust ntop and with a good coverage. All my efforts are heading to that kind of thing. The reason is that ntop is good, but the code base is written in C++03, with the new C++17/++20 we can achieve all that.
Also code reduction, less is more.  So before moving around the things and make other mess, we might want have 
a test suite for avoiding regressions. One of the problems of current ntop developement is a light regression tool.
Now this intent is duplicated in any branch, pr that i am using, if you want to work with me on this, we can set a regular weekly 
call on saturday, at your preferred timetable, create a roadmap and and doing weekly small steps to converge.

giorgiozoppi:
@ouliuquan  my point is that before messing around with code changes we should have a strong regression tool, unit tests, integration tests in order to avoid to introduce bugs. We want to strive for an high quality opensource product.

lucaderi:
Closing for inactivity and because the source code changed in the meantime.

PR #6306, opened 2022-02-13
- Removed an unused variable

giorgiozoppi:
In this PR just removed an unsed  variable warning and moved the declaration of use_json. It's a really minor thing. 
In case you have clickhouse you don't notice the warning in all other cases yes.

- Moved use_json near where it's needed.
	modified:   src/LuaEngineInterface.cpp
	modified:   src/LuaEngineNtop.cpp


giorgiozoppi:
stash this since it's already solved.

PR #6318, opened 2022-02-15
configure.ac.in: Allow dynamic linking against ndpi 3.0

kraj:
Linking statically is problematic since NDPI_LIBS is '-lndpi -lm'
and when we use -Bstatic it also brings in -lm to use libm.a and on some
architectures ( x86 ) which this does not work and results in missing symbols

Fixes
ipe-sysroot/usr/lib/libm.a(e_logf.o): in function `logf_ifunc_selector':
/usr/src/debug/glibc/2.35-r0/git/math/../sysdeps/i386/i686/multiarch/ifunc-sse2.h:30: undefined reference to `_dl_x86_cpu_features'

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/configure.ac.in
+++ b/configure.ac.in
@@ -234,10 +234,8 @@ if test -d /usr/local/include/ndpi ; then :
 fi
 
 PKG_CHECK_MODULES([NDPI], [libndpi >= 2.0], [
-   NDPI_INC=`echo $NDPI_CFLAGS | sed -e "s/[ ]*$//"`
-   # Use static libndpi library as building against the dynamic library fails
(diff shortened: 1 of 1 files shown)


lucaderi:
We prefer to link against the static libndpi library for many reasons. As I would like to accept your PR, please change it so that by default the static library is used, but if you add a new flag to use the shared library I am open to it.

lucaderi:
@kraj Any news?

kraj:
> @kraj Any news?

I have applied the proposed patch and it fixed the issue for yocto, I did have time to implement what you are asking for yet

lucaderi:
Closing for inactivity

PR #6346, opened 2022-02-23
Added secure cookie attributes to ALL cookies on 302 response

gatkisson:
Our vulnerability assessment tools kept identifying ntopng as having vulnerabilities, specifically that cookies were missing the "Secure" attribute and the "httpOnly" attribute.  When working to mitigate the issue we found that the 302 redirect response was not including these attributes on the user or password cookies, but was including them on the session cookie.

We used the same method to set the secure cookie attributes across all 3 cookies which should resolve the issue.

While the original vulnerability does not appear to be exploitable, this change will allow ntopng to more easily pass vulnerability scans.

I've attached 2 images, one that shows the output from the vulnerability scan, and another that shows the response headers.
ntopVulnerabilityScan

ntopResponseHeaders


lucaderi:
Thank you for your contribution. We appreciate your fix.

PR #6519, opened 2022-04-19
Upgrade C++ standard to C++11 standard

ouliuquan:
This option applies to CentOS 7 (GCC 4.8) and above.


lucaderi:
@giorgiozoppi What do you think?

ouliuquan:
actually in CentOS 7 (GCC 4.8), the compiler is supported to C++1y, this is so why I have make misstake in title. In my jobs, I had used the C++1y (std=c++1y) for over 7 years, so we could more than c++11.

ouliuquan:
The point for advanced versions of C++ is that we can use a lot of libraries. There are currently a lot of libraries that need to be basically C++11.

ouliuquan:
Consider to C++1y will be better, new PR have open.

giorgiozoppi:
@ouliuquan   Ntop is not yet ready to pass to modern c++ but we can try to update the compiler to G++ 5 and  clang and see if it breaks something. If current regression in lua and manual testing works we can pass to C++17, just changing the makefile.  Please sync we me if you want work in this sense. There's an ongoing effort to move ntop to C++20.

giorgiozoppi:
@ouliuquan https://github.com/ntop/ntopng/blob/ntopcpp20/CMakeLists.txt Can you also help me on this branch?

ouliuquan:
@giorgiozoppi If we don't plan to support CentOS/RedHat 7 series, above G++5 or Clang is not an issue, but if we do need to support CentOS/RedHat 7, then it won't be an option. But I see that CentOS 7 will have a long life, so it's mainly about the choice of ntop.

Yes, I have seen this branch "ntopcpp20", I can work on this branch and see if I can do something.

PR #6522, opened 2022-04-19
Upgrade C++ standard to C++1y standard

ouliuquan:
This option works on CentOS 7 (GCC 4.8) and above, other Linux OSes have been upgraded to higher GCC, so CentOS 7 needs to be considered, C++1y (before C++14) is what it can do.
The point for advanced versions of C++ is that we can use a lot of libraries. There are currently a lot of libraries that need to be basically C++11.


ouliuquan:
Hi, I've been reconsidering for half a day and I think C++1y would be better for GCC4.8 for CentOS7, the highest C++ standard version it can do. It looks like we will support CentOS 7 for a long time. @lucaderi @giorgiozoppi

lucaderi:
Thank you

giorgiozoppi:
Good at this point you may want to test in centos7 if something breaks,

giorgiozoppi:
when we've already c++11 we can work to replace our batch workpool for launching scripts using std::thread. if you @ouliuquan  want to work on that let me know. Inside the current Ntop there is a ThreadPool that schedule a activities to do. Using C++11/14/17 it will be the perfect candidate towards the code reduction for Ntop. The idea to have a standard thread library in C++ will enable portability to Windows/Mac and other systems supported by libc/libstdc++ with no cost.

ouliuquan:
Yes, I have seen the "[threadpooltest](https://github.com/ntop/ntopng/tree/threadpooltest)" branch and am interested. This is one of the reasons why I plan to upgrade to Modern C++. In this case, we can do more work and in a more standard way when using the standard library. It is also possible to port to Windows/Mac. @giorgiozoppi

giorgiozoppi:
Ok, i will be your servant. How do you think is the best strategy  to proceed? Can we have a shared diagram on what threadpool t does in order to think about that and their interaction with other parts? When we've a shared common design to discuss we can proceed on how make it happen @ouliuquan

ouliuquan:
You are so kind, I just learned about this project for a few weeks and you and the other contributors are my teachers. ntopng is a huge project and I still need to learn more about its architecture. Right now what I'm doing is PR #6298  that you created, it helps me understand the relationship, this is I am plan to do this comment https://github.com/ntop/ntopng/pull/6298#issuecomment-1107350586 , we only do one thing before another finished.  Hopefully I can learn from you faster and do PRs for branches. @giorgiozoppi

giorgiozoppi:
Ok, proceed and if you're stuck i will remove your blockers. If you need to know better things or you don't know just ask. We're here to help. More then happy to review the code

Issue #6150, opened 2021-12-21
Split Local Host Counter


lucaderi:
Traditionally in ntopng local hosts are accounted regardless of their status (sent/rcvd traffic or rcvd-only). This means that a local host that is active because of a remote host scan, is active as a local host that has sent/rcvd data.

This ticket is to create two "local host" counters
- counter of active (i.e. hosts that have made traffic) local hosts
- counter of passive i.e. hosts that have received but not made traffic) local hosts

This counter will be represented in the top menubar and used as an indicator for scanning and used to generate alerts.

PR #6602, opened 2022-05-14
Ticket #6150

nessuno133:
Try to add Split Local Host Counter

### DIFF
--- a/include/ntop_typedefs.h
+++ b/include/ntop_typedefs.h
@@ -1127,9 +1127,9 @@ typedef struct {
 } PeriodicityStatsKey;
 
 typedef struct _localhost_walker{
-  u_int64_t numberActive;
-  u_int64_t numberBroadOrMulti;
-  u_int64_t numberPassive;
+  u_int32_t numberActive;
+  u_int32_t numberBroadOrMulti;
+  u_int32_t numberPassive;
 } localhost_walker;
 
 #endif /* _NTOP_TYPEDEFS_H_ */
--- a/include/ntop_typedefs.h
+++ b/include/ntop_typedefs.h
@@ -1126,4 +1126,10 @@ typedef struct {
   u_int8_t l4_proto;
 } PeriodicityStatsKey;
 
+typedef struct _localhost_walker{
+  int16_t numberActive;
+  int16_t numberBroadOrMulti;
+  int16_t numberPassive;
+} localhost_walker;
+
 #endif /* _NTOP_TYPEDEFS_H_ */
--- a/include/ntop_typedefs.h
+++ b/include/ntop_typedefs.h
@@ -1127,9 +1127,9 @@ typedef struct {
 } PeriodicityStatsKey;
 
 typedef struct _localhost_walker{
-  int16_t numberActive;
-  int16_t numberBroadOrMulti;
-  int16_t numberPassive;
+  u_int64_t numberActive;
+  u_int64_t numberBroadOrMulti;
+  u_int64_t numberPassive;
 } localhost_walker;
 
 #endif /* _NTOP_TYPEDEFS_H_ */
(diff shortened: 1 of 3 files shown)


include/ntop_typedefs.h
```diff
@@ -1126,4 +1126,10 @@ typedef struct {
   u_int8_t l4_proto;
 } PeriodicityStatsKey;
 
+typedef struct _localhost_walker{
+  int16_t numberActive;
```
lucaderi: 16 bit is not enough

src/NetworkInterface.cpp
```diff
@@ -6344,6 +6344,8 @@ void NetworkInterface::lua(lua_State *vm) {
   lua_push_bool_table_entry(vm, "has_seen_external_alerts", hasSeenExternalAlerts());
   lua_push_bool_table_entry(vm, "has_seen_ebpf_events", hasSeenEBPFEvents());
 
+  countActiveLocalHosts(vm, true);
```
lucaderi: This should go under stats (lua table)

src/NetworkInterface.cpp
```diff
@@ -9870,3 +9872,55 @@ int NetworkInterface::walkActiveHosts(lua_State* vm,
...
+  bool isLocal;
+
+  if(host == NULL)
+    return(false);
+
+  isLocal = host->isLocalHost() || host->isSystemHost();
+
+  if(isLocal) {
+  
+    if ((host->getNumPktsSent() == 0) and (host->isBroadcastHost() or host->isMulticastHost())) {
```
lucaderi: Why (host->getNumPktsSent() == 0) ?

src/NetworkInterface.cpp
```diff
@@ -9870,3 +9872,55 @@ int NetworkInterface::walkActiveHosts(lua_State* vm,
...
+      m->numberBroadOrMulti++;
+    } else {
+      if (host->getNumPktsSent()>0 and host->getNumPktsRcvd()>0) {
+        m->numberActive++;
+      } else {
+        m->numberPassive++;
```
lucaderi, line 9893: A host that has sent only traffic can't be passive

src/NetworkInterface.cpp
```diff
@@ -9870,3 +9872,55 @@ int NetworkInterface::walkActiveHosts(lua_State* vm,
...
+  int rc;
+
+  m.numberActive= 0;
+  m.numberBroadOrMulti= 0;
+  m.numberPassive= 0;
+
+  rc = walker(&begin_slot, true, walker_hosts, active_localhosts_walker, (void*)&m);
+
+  lua_push_int32_table_entry(vm, "active", m.numberActive);
+  lua_push_int32_table_entry(vm, "broadormulti", m.numberBroadOrMulti);
```
lucaderi: This data should go under a table

src/NetworkInterface.cpp
```diff
@@ -9870,3 +9872,55 @@ int NetworkInterface::walkActiveHosts(lua_State* vm,
...
+
+  u_int32_t begin_slot = 0;
+  localhost_walker m;
+  int rc;
+
+  m.numberActive= 0;
+  m.numberBroadOrMulti= 0;
+  m.numberPassive= 0;
+
+  rc = walker(&begin_slot, true, walker_hosts, active_localhosts_walker, (void*)&m);
```
lucaderi, line 9915: Can you suggest how to avoid scanning the interface to compute these fields?

lucaderi (changes requested):
Good start, but this is not yet complete.

src/NetworkInterface.cpp
lucaderi: This problem is still pending, namely, we cannot do a full hash scan whenever a page is rendered. So this work is good as a starting point but it needs to be consolidated before merge

lucaderi:
Closing as this code is still incomplete and needs cleanup.

Issue #7126, opened 2023-01-19
Implement VLAN Traffic Check


lucaderi:
This is to request the implementation of a new check that
- allows the user to specify (e.g. a comma-separated list such as 40,47,234) a list of VLANs to supervise
- trigger an alert whenever ntopng observes a flow with bidirectional traffic (i.e. RX > 0 and TX > 0). Flows with TX only, or with local destinations (except multicast and broadcast) should not be considered
- the generated alert should include the details of the flow that triggered such alert

MatteoBiscosi:
Implemented as requested

PR #7207, opened 2023-02-06
Fix VLANs form. (#7126)

NicoMaio:


lucaderi:
Still not working, just cosmetic changes.

Issue #6532, opened 2022-04-21
Flow exporter selection in Live Flows vs selected Observation Point


cardigliano:
When selecting an OP in the dropdown, the Live Flows page still reports the full list of Flow Exporters in the dropdown used to filter by FE. This leads to an empty flows list when selecting a FE which does not belong to the selected OP, and this is useless. It would be nice to show only the FE in the selected OP.

cardigliano:
Implemented in PR #7230

PR #7230, opened 2023-02-13
Update FE dropdown menu entries. (#6532)

NicoMaio:
Now on the FE filter dropdown are present only the IPs of the observation point selected.


cardigliano:
I confirm this is working

Issue #7346, opened 2023-03-25
Add RIPEstat link to Host Details page


lumarel:
I have been recently using the new RIPEstat launchpad a whole lot more because it consistently delivers a lot of interesting information.

**What would you like to add or change?**:
Add a RIPEstat Lookup button beside the Whois Lookup button in the Host Details page.
Also I noticed that the AS lookup in the list view does not work anymore, and the one on the AS Details page already uses RIPEstat.

**Why do you and others need this?**:
It's a useful source of information for external IPs, AS networks and domains.

I also already played around with a test environment and created a patch for it, I can send a PR for that if wanted.
![image](https://user-images.githubusercontent.com/42647570/227738931-7fdb2335-a996-4ea4-9a58-3cc782897412.png)

MatteoBiscosi:
Hi @lumarel you have have a pr to fix this issue, sure thing, you can open it, this is an open source project so we gladly welcome pr

lumarel:
@MatteoBiscosi Okay thank you!

PR #7350, opened 2023-03-27
Add RIPEstat link to Host Details page

lumarel:
As mentioned in #7346 I would like to add a RIPEstat lookup link to the host details page.
I created a translation for every language, by looking what was already there for Whois and looking it up on a translator, so it might not be completely correct (beside my main languages English and German).

Beside that I noticed that the robtex link looks to be broken and already using RIPEstat on the details page, so this also changes the list link.

I tested this change on a running system, so it shouldn't break something.


MatteoBiscosi:
Hi @lumarel , thanks a lot for this contribution, everything seems perfect. Also, yes, I didn't notice that the robtex link is not working anymore, thanks for fixing it

PR #7372, opened 2023-04-06
Add code formatter in the CI pipeline

patacca:
This workflow will use clang-format to enforce a coding style (defined in `.clang-format`) to all the cpp source files.

The coding style is based on [Google style](https://google.github.io/styleguide/cppguide.html) with few differences:
 - Indentation is 2 spaces
 - Do not reorder the included headers because this potentially disrupt the dependencies between them

**Note:** This PR will affect ~600 files, in order to preserve the git blame it's possible to put the commit ID in which the code formatting takes place in the `.git-blame-ignore-revs` file, this will force git blame to ignore these changes (https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revltrevgt)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,2 @@
+BasedOnStyle: Google
+
--- a/.clang-format
+++ b/.clang-format
@@ -1,2 +1,7 @@
+---
 BasedOnStyle: Google
-
+---
+Language: Cpp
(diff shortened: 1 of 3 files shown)


lucaderi:
Thank you

Issue #7396, opened 2023-04-13
[nEdge] Make Multicast repeater configurable


cardigliano:
A multicast repeater has been already implemented in nEdge to forward MDNS traffic. This can be currently enabeld under the Misc menu in the System Setup page.
It is requested to add a configuration page with a table similar to the Inter-LAN Rules page in order to let the user configure a list of:

- Multicast IP
- Port
- Subset of the LAN interfaces where this should be enabled

An instance of the MDNSRepeater class has to be instantiated for each configured entry in order to handle the forwarding of multiple multicast services. Using this page, the MDNS forwarder can be configured (as a preset) using the IP and port currently hard coded in the MDNSRepeater class.

PR #7409, opened 2023-04-17
Add multicast forwarders. (#7396)

NicoMaio:
We need to update the dist after the merge.


src/Ntop.cpp
```diff
@@ -612,24 +614,62 @@ void Ntop::start() {
...
 
-  if (is_mdns_enabled) {
-    mdnsRepeater = new (std::nothrow) MDNSRepeater();
 
-    if (mdnsRepeater == NULL)
-      ntop->getTrace()->traceEvent(TRACE_ERROR,
-                                   "Error occured instantiating MDNSRepeater");
-    else
-      mdnsRepeater->start();
+    MulticastForwarder *multicastForwarder = new (std::nothrow) MulticastForwarder(!strcmp(type.c_str(),"MDNS") , ip, stoi(port), interfaces);
```
cardigliano: Do we really need to check if the type is "MDNS" here?

src/Ntop.cpp
```diff
@@ -612,24 +614,62 @@ void Ntop::start() {
...
+    multicastForwarders.push_back(multicastForwarder);
+    
+    memset(key, 0, sizeof(key));
+    memset(repeater, 0, sizeof(repeater));
```
cardigliano: Why are you resetting repeater here? I would move this before redis->get

src/Ntop.cpp
```diff
@@ -612,24 +614,62 @@ void Ntop::start() {
...
 
-    if (mdnsRepeater == NULL)
-      ntop->getTrace()->traceEvent(TRACE_ERROR,
-                                   "Error occured instantiating MDNSRepeater");
-    else
-      mdnsRepeater->start();
+    MulticastForwarder *multicastForwarder = new (std::nothrow) MulticastForwarder(!strcmp(type.c_str(),"MDNS") , ip, stoi(port), interfaces);
+    multicastForwarders.push_back(multicastForwarder);
+    
+    memset(key, 0, sizeof(key));
```
cardigliano: Why are you resetting key here? It seems it's useless

src/Ntop.cpp
```diff
@@ -612,24 +614,62 @@ void Ntop::start() {
...
-  if (is_mdns_enabled) {
-    mdnsRepeater = new (std::nothrow) MDNSRepeater();
 
-    if (mdnsRepeater == NULL)
-      ntop->getTrace()->traceEvent(TRACE_ERROR,
-                                   "Error occured instantiating MDNSRepeater");
-    else
-      mdnsRepeater->start();
+    MulticastForwarder *multicastForwarder = new (std::nothrow) MulticastForwarder(!strcmp(type.c_str(),"MDNS") , ip, stoi(port), interfaces);
+    multicastForwarders.push_back(multicastForwarder);
```
cardigliano: A check on the multicastForwarder allocation is missing, please always check that "new" allocates the object

src/Ntop.cpp
```diff
@@ -612,24 +614,62 @@ void Ntop::start() {
...
+        } else if (h == 1) {
+          ip = token;
+        } else if (h == 2) {
+          port = token;
+        } else if(h == 3) {
+          interfaces = token;
+        }
+        token = strtok(NULL, "|");
+        h++;
+    }
```
cardigliano, line 654: Please check that all the required options have been set

PR #7428, opened 2023-04-28
Add search map on aggregated live flows.

NicoMaio:
Add filter on aggregated live flows using the map search parameter.

### DIFF
--- a/src/NetworkInterface.cpp
+++ b/src/NetworkInterface.cpp
@@ -11205,125 +11205,110 @@ void NetworkInterface::build_lua_rsp(lua_State *vm, AggregatedFlowsStats *flow_s
 
 /* **************************************************** */
 
-/* Verify function for the map search filter */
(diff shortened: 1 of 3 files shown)


src/NetworkInterface.cpp
```diff
@@ -11316,9 +11467,6 @@ void NetworkInterface::sort_flow_stats(
 
   }
```
lucaderi, line 11469: Why you have removed memory free?

src/NetworkInterface.cpp
```diff
@@ -11262,7 +11407,13 @@ void NetworkInterface::sort_flow_stats(
   } else {
     std::unordered_map<string, AggregatedFlowsStats *>::iterator it;
     for (it = count_info->begin(); it != count_info->end(); ++it) {
-      vector.push_back(it->second);
+      if (map_search && strcmp(map_search, "") != 0) {
```
lucaderi, line 11410: Missing search on IP addresses

lucaderi (commented):
Not bad but there are issues to solve

PR #7430, opened 2023-04-30
Add first fuzzing harness

patacca:
Implement a first rudimentary fuzzing harness to fuzz `NetworkInterface::dissectPacket()`
It works with libfuzzer (optionally with libprotobuf) and AFL++

Refer to README.md for more informations on how to build and run a fuzzing instance


lucaderi:
I will accept this PR that look good to us. we just need to automate the commands as in the current state it is a bit complicated to build the code.

PR #7443, opened 2023-05-04
Fix null pointer dereference

patacca:
This will fix the null pointer dereference error caused by this input [crash.pcap.zip](https://github.com/ntop/ntopng/files/11394243/crash.pcap.zip)

There also is a little bit of coding style formatting.

### DIFF
--- a/src/NetworkInterface.cpp
+++ b/src/NetworkInterface.cpp
@@ -2991,9 +2991,8 @@ decode_packet_eth:
       if (ntop->getPrefs()->do_ignore_macs()) ethernet = &dummy_ethernet;
 
       if (ethernet == NULL) {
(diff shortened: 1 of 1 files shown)


cardigliano:
@patacca please note that the "Fix crash" commit also chances a lot the formatting, making it difficult to identify the real change. Please split the commits next time (note that the "Fix formatting" commit instead changes just a couple of lines).
In this PR, it is enough to comment the line you changed to dix the null pointer to let us review it. Thank you.

cardigliano:
This was the actual fix:
https://github.com/ntop/ntopng/pull/7443/files#diff-57758010ff492d76bc022b034aefa549028041e56cdc4f39a024c2cbb9b90539R2993

Issue #7442, opened 2023-05-03
Split Protocols (Guess and not Guess)


lucaderi:
In
![image](https://user-images.githubusercontent.com/4493366/235991574-428f7fa6-7f8c-496b-acca-89a024dea433.png)

for each application protocol (e.g. TLS) if you have traffic for TLS and TLS (guess) then these do not have to be merged as today, but be split into two protocols.

Note that inside the flow it's written if the flow protocol is guessed or not

lucaderi:
An application protocol is guessed as returned by getConfidence(). In case the confidence is
   case NDPI_CONFIDENCE_DPI_CACHE:
    case NDPI_CONFIDENCE_DPI_PARTIAL_CACHE:
    case NDPI_CONFIDENCE_DPI:
    case NDPI_CONFIDENCE_NBPF:
 
then is it NOT guessed, otherwise is guessed.

MatteoBiscosi:
Implemented as requested

PR #7445, opened 2023-05-04
Update with guessed and dpi protocols. (#7442)

NicoMaio:
Update the aggregated live flows page on application protocol criteria.


src/NetworkInterface.cpp
```diff
@@ -10924,10 +10924,12 @@ bool NetworkInterface::compute_protocol_flow_stats(GenericHashEntry *node,
   std::unordered_map<u_int64_t, AggregatedFlowsStats *> *count =
       static_cast<std::unordered_map<u_int64_t, AggregatedFlowsStats *> *>(
           user_data);
+  u_int64_t is_not_guessed = (u_int64_t) int(f->isDPIDetectedFlow());
 
   /* <0 (16 bit)><vlan_id (16 bit)><app_protocol (16 bit)><master_protocol (16
    * bit) */
   key = ((u_int64_t)vlan_id << 32) +
+        ((u_int64_t) is_not_guessed << 32) +
```
lucaderi: 1. update the comment on line 10929
2. the shift to is_not_guessed should be 48 and not 32

lucaderi (changes requested):

PR #7453, opened 2023-05-08
Copyright fix for Y2K

todb:
This PR fixes some nonsense ranges mentioned in copyright strings, like:

```
(C) 1998-23 ntop.org
```

It's unclear how one is supposed to parse "1998-23" since the two most obvious are "1998 minus 23 (1975)" and "1998 - 1923" (when in fact it's certainly supposed to be parsed as "1998-2023," but computers are dumb and can't figure that out, thus the whole [Y2k problem](https://education.nationalgeographic.org/resource/Y2K-bug/)).

But, regardless of the weird YYYY-YY range, copyright statements generally should merely be the date of first published, per [this helpful site](https://www.copyrightlaws.com/copyright-symbol-notice-year/#:~:text=Which%20Year%20Should%20You%20Use,first%20publication%20of%20the%20work.). [Here's the real law](https://www.copyright.gov/title17/92chap4.html) on copyright in the US (note, no ranges). The Free Software Foundation recommends ranges, but the FSF is wrong and should be ignored generally anyway. That last part is like, just my opinion, man.

I also fixed a couple other ranges when they were nearby, but the main one I wanted to fix was the one emblazoned at the bottom of every ntop screen when I'm using it. I can comfortably ignore the other nonsense copyright ranges scattered around license text and source code, but that one I have to look at all the time, and it annoys me.

Finally, I'll be damned if I leave a Y2K bug unfixed this late in my career. I did all this almost a quarter century ago.

### DIFF
--- a/attic/tools/nindex_export_to_ch.sh
+++ b/attic/tools/nindex_export_to_ch.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-#  Copyright (C) 1998-22 - ntop.org
+#  Copyright (C) 1998 - ntop.org
 #
 #  http://www.ntop.org/
 #
--- a/include/ntop_defines.h
+++ b/include/ntop_defines.h
@@ -1,6 +1,6 @@
 /*
  *
- * (C) 2013-23 - ntop.org
+ * (C) 2013 - ntop.org
  *
  *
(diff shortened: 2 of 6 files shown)


lucaderi:
We have discussed this internally and decided that we'll leave as it until we take a final decision on the subject. Thanks for the PR

PR #7462, opened 2023-05-13
Add CI for fuzzing

patacca:
Enable the CI for fuzzing with oss-fuzz https://github.com/google/oss-fuzz/pull/10218

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,28 @@
+name: CIFuzz
+on: [push, pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        sanitizer: [address]
+    steps:
+    - name: Build Fuzzers - ${{ matrix.sanitizer }}
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'ntopng'
+        sanitizer: ${{ matrix.sanitizer }}
+    - name: Run Fuzzers - ${{ matrix.sanitizer }}
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'ntopng'
+        sanitizer: ${{ matrix.sanitizer }}
+        fuzz-seconds: 1200
+    - name: Upload Crash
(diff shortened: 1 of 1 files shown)


IvanNardi:
Judging from the logs, it seems to me that the fuzzer didn't run at all. Is that somehow expected?

```
2023-05-13 11:08:03,442 - root - INFO - Starting fuzzing
Fuzzing logs:
13/May/2023 11:08:03 [Prefs.cpp:2289] ERROR: Unable to locate docs dir
13/May/2023 11:08:03 [Ntop.cpp:401] ERROR: Invalid directory install specified
2023-05-13 11:08:03,893 - root - INFO - Fuzzer: fuzz_dissect_packet. Detected bug.
2023-05-13 11:08:03,894 - root - INFO - Trying to reproduce crash using: /tmp/tmp12c6yf8w/tmp5jizwzci.
2023-05-13 11:08:03,923 - root - INFO - Reproduce command returned: 255. Reproducible on /github/workspace/build-out/fuzz_dissect_packet.
2023-05-13 11:08:03,924 - root - INFO - Crash is reproducible.
```

patacca:
I didn't notice that issue, at first I assumed the error was due to the unavailability of the corpus (ultimately caused by [this](https://github.com/google/oss-fuzz/issues/10323)) but this seem to be a different and rather important issue.
I guess the problem is that the directory from which is executed the fuzzer is different from the one where the actual binary is located.
The need for the docs directory at run time (as well as the others) is definitely a bother, unfortunately it's not easy to remove them from the code. It might be a better idea to create them directly in the fuzzing target, in the `LLVMFuzzerInitialize` function.

IvanNardi:
@patacca, definitely better; now the fuzzer run. There are some leaks, but these issues are independent from this specific PR

patacca:
I think this one can be merged now

IvanNardi:
@patacca, @lucaderi  not sure if you noticed that, but there are errors in the CI fuzzing job
Most of them are the "standard" heap-buffer-overflow stuff:
https://github.com/ntop/ntopng/actions/runs/5176514953/jobs/9325418664
https://github.com/ntop/ntopng/actions/runs/5178687974/jobs/9330505856
https://github.com/ntop/ntopng/actions/runs/5179881064/jobs/9333350925

but this one is a timeout (https://github.com/ntop/ntopng/actions/runs/5168472335/jobs/9310092458) and it might be related to the issue seen in oss-fuzz...

Issue #7438, opened 2023-05-03
Extend timeseries driver:listSeries


lucaderi:
In driver:listSeries(schema, tags_filter, wildcard_tags, start_time) add a new parameter called end_time and make sure all the calls to the above function populate it

MatteoBiscosi:
Implemented as requested

PR #7466, opened 2023-05-16
Handle end time on timeseries. (#7438)

NicoMaio:
Add end time on time-series.


lucaderi:
What is this PR about?

NicoMaio:
It's about #7438.

Issue #7493, opened 2023-05-22
Implement Ports Analysis Page


NicoMaio:
A new page must be implemented under the Hosts menu's entry to see which hosts are using the selected server port. 
On this page will be present two tabs: Live and Historical. 
The live analysis page will present two dropdown menus to select the protocol (for example, UDP or TCP) and a second dropdown menu with the UDP's or TCP's ports.

After the port is selected, the page will be shown server hosts using the port.

MatteoBiscosi:
Implemented as requested

PR #7498, opened 2023-05-25
Add Server Ports Analysis page. (#7493)

NicoMaio:
Add a new Local Hosts Server Ports Analysis page after the merge is mandatory to merge the related PR on the pro repo and build the dist.


src/HostsPortsAnalysis.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include "ntop_includes.h"
+
+
+/* **************************************************** */
+
+void HostsPortsAnalysis::add_host_details(HostDetails * host_details_to_insert) {
+    std::unordered_map<u_int64_t, HostDetails *>::iterator it;
+
+    it = hosts_details.find(host_details_to_insert->get_host_key());
```
MatteoBiscosi, line 32: Please check that host_details_to_insert isn't NULL before using it, otherwise it could bring to a SEGV

PR #7500, opened 2023-05-27
Ignore leaks

patacca:
Disabling leaks detection should be useful for increasing the coverage.
See also https://github.com/google/oss-fuzz/issues/10323


lucaderi:
Thank you

Issue #7538, opened 2023-06-08
Periodicity Table: Add Duration


lucaderi:
In the periodicity table add duration column that displays since when the communication is in place (in essence difference last-first  seen)

MatteoBiscosi:
Verified as fixed

Issue #7537, opened 2023-06-08
Missing Drill Down in Periodicity/Asset/Service Table


lucaderi:
In all tables of Periodicity/Asset/Service add (if Clickhouse is enabled) the ability to drill down to flows in the historical database

MatteoBiscosi:
Verified as fixed

PR #7548, opened 2023-06-12
Add historical flows links and duration column in periodicity table.(#7537) (#7538)

NicoMaio:
It needs to check on the pro repository for the C++ code update after the merge is necessary to build the dist.


lucaderi:
@NicoMaio The PR contains ; in the URL that is not correct. Please fix them

Issue #7604, opened 2023-06-29
Unexpected Host Name Resolution Flapping


lucaderi:
As host serialization has been removed from ntopng, every time a host is created in memory the name resolution will start over. So it can be that at time X the host has no name, then at the X+1 it is resolved. Historical flows dumped before X+1 will have a numerical name, and others will have a symbolic name. If the host appears/disappears from memory (i.e. it does not make constant traffic) looking at historical flows, you will see records symbolic and other numeric.
This ticket is to request to serialize the hostname resolution (and not all the stats as it used to be) in a Redis key with 1day TTL so that we can avoid the problem

PR #7605, opened 2023-06-29
Fix localhost name resolution. (#7604)

NicoMaio:
Fix localhost name resolution setting hostname taken from Redis.


src/LocalHost.cpp
```diff
@@ -117,8 +117,12 @@ void LocalHost::initialize() {
         (isIPv6() &&
          ((strncmp(strIP, "ff0", 3) == 0) || (strncmp(strIP, "fe80", 4) == 0))))
       ;
-    else
+    else {
       ntop->getRedis()->getAddress(strIP, rsp, sizeof(rsp), true);
+      if(strcmp(rsp, "")) {
```
lucaderi, line 122: if(rsp[0] != '\0') is more efficient

PR #7756, opened 2023-08-11
fuzz: force fuzzing job on CI to fail if it founds any errors

IvanNardi:
Pro: you can easily detect if you have errors in your code base.

Cons: the CI might fail for unrelated errors, i.e. for errors unrelated to the code you are trying to push/commit/merge. And that *will* happen a lot, especially just after this PR has been merged, because there are lots of errors right now.

Memory leak detection is still disabled.

This is the behavior that we have on nDPI too.


IvanNardi:
Not working as expected...

PR #7848, opened 2023-09-27
Add mining currency in flow info column.

NicoMaio:
To display in the info column the cryptocurrency when there's a flow with a miner.


lucaderi:
I will fix the missing free

lucaderi:
https://github.com/ntop/ntopng/commit/4e266a23ff0113c110297b062e2b13e7645768f7

Issue #7917, opened 2023-10-17
DataTable Resets Page When Data is Refreshed


lucaderi:
Take a data table and move to page X (where X > 1) then wait until refresh happens or click on the refresh icon. The page is set to 1 instead of staying at page X

lucaderi:
Verified as fixed

PR #7920, opened 2023-10-17
Fix change table page on refresh. (#7917)

NicoMaio:
The table will remain on the current page, whether refreshed manually by clicking the icon or automatically.

UPDATE the 'dist' directory after merging and initiating the builds.

### DIFF
--- a/http_src/vue/select_table_page.vue
+++ b/http_src/vue/select_table_page.vue
@@ -114,7 +114,9 @@ function back_page() {
 }
 
 function change_active_page(new_active_page, new_start_page_button) {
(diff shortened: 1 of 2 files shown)


lucaderi:
Probably it's better to divide the problem in 2: 
- autorefresh with no line change (the number of table items doesn't change, just the content)
- manual refresh (lines can decrease)

In the first case (new number of lines >= old number of lines) if we do not reset the pageId we have an inconsistent tabe. In the second case there is no need to reset the Id.

For VS we can avoid resetting the pageId for sure, in other cases we need to find a better strategy.

PR #8200, opened 2024-02-01
Fix L7 Protocol usage & empty table statement using the view interface in Server Ports Analysis page

NicoMaio:
Fix the similarity in content between the table and chart view by adding the L7 master and app protocol filter.
Fix the empty table view in the case of the view interface. 

Please also see the pull request on the PRO repository, and after the merge, please build and push the DIST.


src/NetworkInterface.cpp
```diff
@@ -11759,21 +11757,15 @@ bool NetworkInterface::get_host_ports(GenericHashEntry *node,
...
-  /* Server Host attribute */
-  u_int16_t srv_vlan_id           = server->get_vlan_id();
-
   /* check l4_proto filter */
   if (flow_l4_proto != filter_l4_proto)
     return (false); /* false = keep on walking */
 
   /* check vlan filter */
-  if (filter_vlan_id != 0) {
+  if (filter_vlan_id != (u_int16_t)-1) {
```
lucaderi: Please #define NO_VLAN (u_int16_t)-1

src/NetworkInterface.cpp
```diff
@@ -11873,9 +11868,33 @@ bool NetworkInterface::get_hosts_by_port(GenericHashEntry *node,
   if ((filter_l4_proto != f->get_protocol()) || (filter_srv_port != f->get_srv_port()))
     return(false);
 
+  ndpi_protocol detected_protocol = f->get_detected_protocol();
```
lucaderi, line 11998: Please use integer arithmetics for comparison

lucaderi (changes requested):

src/NetworkInterface.cpp
NicoMaio: Fixed by d77d5568d339619566e97dc7302fe609fe36656e

src/NetworkInterface.cpp
NicoMaio: Fixed by e52ef99bbbc0fd47be3ae41a14b232e7b3f60ec3

include/HostsPortsAnalysis.h
```diff
@@ -26,11 +26,12 @@
 
 class HostsPortsAnalysis {
 
-private:
-  u_int16_t port = 0;
-  int l7_protocol_id = 0;
-  u_int8_t l4_protocol_id = 0;
-  std::unordered_map<u_int64_t, HostDetails *> *hosts_details;
+    private:
+        u_int16_t port = 0;
```
lucaderi, line 30: this is a bad practice as you should set values in constructor

Issue #8347, opened 2024-04-22
Counter Overflow Bug


lucaderi:
<img width="611" alt="image" src="https://github.com/ntop/ntopng/assets/4493366/a1acc103-715e-4b76-b400-0570f49cceb1">

lucaderi:
Verified as fixed

PR #8349, opened 2024-04-22
Possible fix for counter overflow (#8347)

NicoMaio:
Possible fix for issue #8347: adding a check on the increment and decrement methods of the counter.


lucaderi:
This does PR not seem to solve the problem

PR #8402, opened 2024-05-24
FIrst integration of check on open server ports

lucaferret:


src/Flow.cpp
```diff
@@ -1246,7 +1245,7 @@ void Flow::endProtocolDissection() {
     updateProtocol(ndpi_detection_giveup(iface->get_ndpi_struct(), ndpiFlow,
                                          &proto_guessed));
     setRisk(ndpi_flow_risk_bitmap | ndpiFlow->risk);
-    setProtocolDetectionCompleted(NULL, 0);
+    setProtocolDetectionCompleted(NULL, 0, 0);
```
lucaderi: Replace 0 with iface->getTimeLastPktRcvd()

src/Flow.cpp
```diff
@@ -1257,7 +1256,7 @@ void Flow::endProtocolDissection() {
 /* Manually set a protocol on the flow and terminate the dissection. */
 void Flow::setDetectedProtocol(ndpi_protocol proto_id) {
   updateProtocol(proto_id);
-  setProtocolDetectionCompleted(NULL, 0);
+  setProtocolDetectionCompleted(NULL, 0, 0);
```
lucaderi: Replace 0 with iface->getTimeLastPktRcvd()

src/LocalHost.cpp
```diff
@@ -632,6 +632,17 @@ void LocalHost::setRxOnlyHost(bool set_it) {
 
 /* ***************************************************** */
 
+void LocalHost::setServerPort(bool isTCP, u_int16_t port, ndpi_protocol *proto, time_t when) {
+  bool set_port_status = usedPorts.setServerPort(isTCP, port, proto);
+  u_int32_t learning_period = ntop->getPrefs()->get_contacted_server_port_learning_period();
```
lucaderi: Move it two lines down inside the if(..)

src/Prefs.cpp
```diff
@@ -39,6 +39,7 @@ extern "C" {
 Prefs::Prefs(Ntop *_ntop) {
   if(trace_new_delete) ntop->getTrace()->traceEvent(TRACE_NORMAL, "[new] %s", __FILE__);
   
+  contacted_server_port_learning_period = 300;
```
lucaderi: Do not hardcode values. Set a #define in ntop_defines.h

lucaderi (changes requested):
Please fix the reported issues

PR #8430, opened 2024-06-04
FreeBSD: fix "attempt to concatenate a nil value" in lua script

coderonline:
After a fresh installation of ntop under FreeBSD 14 in a jail I got an error message on several pages, e.g. `https://ntop.exmaple.com/lua/if_stats.lua` saying something about `os.tmpname` and/or `nil` value. Since `os.tmpname` should better not be used anyways and a check for an empty `f_name` prevents the function from trying to delete a non-existing file,  `io.tmpfile` can simply be used and works for me.

### DIFF
--- a/scripts/lua/modules/os_utils.lua
+++ b/scripts/lua/modules/os_utils.lua
@@ -82,13 +82,7 @@ function os_utils.execWithOutput(c, ret_code_success)
    
    if(debug) then tprint(c) end
 
-   if(is_freebsd) then
-      f_name = os.tmpname()
-      os.execute(c.." > "..f_name)
-      f = io.open(f_name, 'r')
-   else
-      f = io.popen(c, 'r')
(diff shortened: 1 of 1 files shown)


cardigliano:
@coderonline with this patch the command (c) is not executed, am I missing something?

coderonline:
That is true, but it did not matter, because the command was `which ntop` and it returned nothing, because the environment does not contain `/usr/local/bin` (_in fact os.environ is nil_). However, no special treatment for FreeBSD is required there. Do you agree?

cardigliano:
os_utils.execWithOutput() is a generic function used to execute commands

coderonline:
True, but this code is within `os_utils.execWithOutput()`

cardigliano:
Sorry I misread the diff, looks good then

PR #8740, opened 2024-10-02
Fix False SYNFLOOD alerts

sulaimansuhas:

Describe changes:
Keeping count of half opened TCP connections instead of SYN packets to  keep track of SYN-FLOOD attacks.


src/Flow.cpp
```diff
@@ -8745,19 +8773,39 @@ MinorConnectionStates Flow::calculateConnectionState(bool is_cumulative) {
...
+      (isTCPFlagSet(dst2src_tcp_flags,TH_SYN)) &&
+      (isTCPFlagSet(dst2src_tcp_flags,TH_ACK)) &&
+      !(isTCPFlagSet(src2dst_tcp_flags,TH_ACK))&&
+      ((isTCPFlagSet(dst2src_tcp_flags,TH_RST))|| 
+      (isTCPFlagSet(src2dst_tcp_flags,TH_RST)) ||
+      (isTCPFlagSet(src2dst_tcp_flags,TH_FIN)) ||
+      (isTCPFlagSet(dst2src_tcp_flags,TH_FIN)))) {
+      updateSynFloodAlerts(false);
+  }
+
```
MatteoBiscosi, line 8808: The code wirtten in this way it's not very clean and hard to understand in the future when debugging. It would have been cleaner simply running a switch case at the end of the calculateConnectionState() function taking account for all the possible connection states and based on the value simply increase or decrease the counter.

src/Flow.cpp
```diff
@@ -8733,6 +8740,27 @@ bool Flow::checkS1ConnState() {
 				   !(isTCPFlagSet(src2dst_tcp_flags,TH_RST)) && !(isTCPFlagSet(dst2src_tcp_flags,TH_RST))     /* NO RST */
 				   ));
 }
+/* **************************************************** */
+
+void Flow::updateSynFloodAlerts(bool connection_opened){
+    if((current_c_state != S0 && !connection_opened) || 
+    (current_c_state !=MINOR_NO_STATE && connection_opened))
+      return;
```
MatteoBiscosi, line 8748: If handled correctly in the calculateConnectionState() function, i don't think this check is needed

src/AlertCounter.cpp
```diff
@@ -67,7 +67,15 @@ void AlertCounter::inc(time_t when, AlertableEntity *alertable) {
...
 }
+/* *************************************** */
+
+void AlertCounter::inc_no_time_window(){
+  current_hits++;
+}
 
+void AlertCounter::dec(){
+  current_hits--;
+}
```
MatteoBiscosi, line 78: Here you have to check that current_hits is higher than 0.
It can happen in many ways that the first packet of a captured flow does not starts as a normal connection (for instance the packet is forged and starts with a SYN and FIN set or simply the flows in captured half-way during the real connection) and this might lead to a counter overflow issue.

src/Flow.cpp
sulaimansuhas: That makes sense! However, I am  a bit confused by the states we have defined as I feel they do not cover all possible connection states, such as when a connection is cancelled after a the server sends the SYNACK packet. Am I missing something or would it be ok to define another connection state?

src/Flow.cpp
sulaimansuhas: These checks are mainly for two cases:

1.  Every data packet sent after the handshake trigger the checkS1ConnState(),
leading to unnecessarily decrementing the count.
2. If SYN packets are being resent from a client we will increment the count unnecessarily.

This logic could be in the calculateConnectionState() function but I thought it would be cleaner here.

src/AlertCounter.cpp
sulaimansuhas: The check in the updateSynFLoodAlerts() function should ensure that we only
decrement the count if the connection has previously been opened. So half captured flows and spoofed packets shouldn't be able to trigger this and cause an overflow I think.

lucaderi:
We believe this code can work but it's a bit complex in its current state.

PR #8747, opened 2024-10-05
Invalid SYNFlood Error Fix

YellowMan02:

Describe changes:
To determine whether a host is an attacker or a victim of a SYNFlood attack, we can count the difference between the number of established and active TCP flows, as a client and as a server (where established means that the 3wh has been performed correctly and active means that the 3wh has been started but not yet completed).


include/Host.h
```diff
@@ -90,6 +90,11 @@ class Host : public GenericHashEntry,
     u_int32_t fin_recvd_last_min, finack_sent_last_min; /* (victim) */
   } fin_scan;
 
+  struct {
+    std::atomic<u_int32_t> num_active_tcp_flows_as_client, num_established_tcp_flows_as_client; /* (attacker) */
```
lucaderi, line 94: These might not be necessary, we'll removed them if so

src/Host.cpp
```diff
@@ -1502,13 +1494,34 @@ void Host::incNumFlows(time_t t, bool as_client) {
 void Host::decNumFlows(time_t t, bool as_client) {
   if (as_client) {
     num_active_flows_as_client--;
+    syn_flood.num_active_tcp_flows_as_client--;
+    syn_flood.num_established_tcp_flows_as_client--;
   } else {
     num_active_flows_as_server--;
+    syn_flood.num_active_tcp_flows_as_server--;
+    syn_flood.num_established_tcp_flows_as_server--;
```
lucaderi: In this method you need to decrease after checking if the counter was previously increased otherwise your lice will break. Please fix it

lucaderi (changes requested):
Almost working, please fix the issue reported.

MatteoBiscosi (approved):

PR #8759, opened 2024-10-21
Deleted atomic variables for host syn flood counters

YellowMan02:

- [ x ] I have signed the ntop Contributor License Agreement at https://github.com/ntop/legal/blob/main/individual-contributor-licence-agreement.md
- [ x ] I have read the contributing guide lines https://github.com/ntop/ntopng/blob/dev/CONTRIBUTING.md
- [ x ] I have updated the documentation (in doc/src/) to reflect the changes made (if applicable)

Describe changes:
Instead of decreasing the counters every time a flow is deleted, reset the syn_flood counters every minute when the syn_flood check is called. 
This way, the logic remains the same and the counters are only updated by the same thread.

### DIFF
--- a/include/Host.h
+++ b/include/Host.h
@@ -94,8 +94,8 @@ class Host : public GenericHashEntry,
   } fin_scan;
 
   struct {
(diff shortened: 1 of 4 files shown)


lucaderi:
Code is not properly fixing the issue. Will be recoded

PR #8864, opened 2024-12-05
Alert remote access fix

YellowMan02:

immagine
minor fix to alert logic


scripts/lua/modules/alert_definitions/flow/alert_remote_access.lua
```diff
@@ -47,7 +47,7 @@ end
 function alert_remote_access.format(ifid, alert, alert_type_params)
    local now = os.time()
    local tstamp_end = alert["tstamp_end"] or now
-   local time = tstamp_end - alert["tstamp"]
+   local time = now - tstamp_end
```
MatteoBiscosi, line 50: This is wrong, the time in the Remote Access alert is used to print the duration of the remote access session, which was correctly calculated before

PR #8926, opened 2025-01-29
Fixes mdns_name handling by escaping single quotes in asset values

DropB1t:

Describe bug:
SQL assets insertion yield an error when inside a "mdns_name" field value is present a single quote
image


MatteoBiscosi:
Thanks for the report of the issue, however the pr does not work entirely because in case the single quote is present on other fields, the issue persists, so i implemented in the lua script the way to clean all the fields from the single quote https://github.com/ntop/ntopng/commit/4d092ae7783973cc41d539fbcba5f4a242cc4d43

Issue #9398, opened 2025-07-16
Dump historical flows to archive before harvesting


lucaderi:
This is to request to implement an optional dump feature, to save historical flows to an archive (e.g. in compresse text or importable format) so that we can store them on a secondary disk for compliancy

lucaderi:
Have a look at https://clickhouse.com/docs/guides/developer/ttl if it can help

PR #9427, opened 2025-07-23
Added ch flows export to file before ttl delete #9398

DGabri:

- [X] I have signed the ntop Contributor License Agreement at https://github.com/ntop/legal/blob/main/individual-contributor-licence-agreement.md
- [X] I have read the contributing guide lines https://github.com/ntop/ntopng/blob/dev/CONTRIBUTING.md
- [X] I have updated the documentation (in doc/src/) to reflect the changes made (if applicable)

Describe changes:

Added ch flows export to file before ttl delete #9398 need to implement path pref and getter in scripts/lua/modules/clickhouse_retention.lua


include/ntop_defines.h
```diff
@@ -790,6 +790,11 @@
 #define DEFAULT_HTTP_INDEX_PAGE \
   ""
 
+#define CONST_PREFS_CH_EXPORT_PATH \
+  NTOPNG_PREFS_PREFIX ".export_flows_path"
```
cardigliano: The key is a bit too generic, I suggest a more descriptive one, e.g. clickhouse_flows_archive_path

include/Prefs.h
```diff
@@ -587,6 +588,9 @@ class Prefs {
   inline char *getHttpIndexPage() {
     return http_index_page;
   };
+  inline char *getCHExportPath() {
```
cardigliano: I suggest "archive" rather than export, which may be misleading

scripts/locales/en.lua
```diff
@@ -7630,7 +7630,7 @@ local lang = {
     ["toggle_dump_pcap_to_clickhouse_title"] = "Dump Pcap Interfaces Flows",
     ["toggle_export_flows_to_archive_description"] = "Save data to the specified archive folder before it's automatically deleted by the retention policy",
     ["toggle_export_flows_to_archive_title"] = "Archive Flows Before TTL Deletion",
-    ["path_export_flows_to_archive_description"] = "Specify The Path To Export Flows To",
+    ["path_export_flows_to_archive_description"] = "Specify The folder Path To Export Flows To: i.e. /var/log/",
```
cardigliano: Text case is mixed

include/ntop_defines.h
```diff
@@ -788,6 +790,11 @@
 #define DEFAULT_HTTP_INDEX_PAGE \
   ""
 
+#define CONST_PREFS_CH_EXPORT_PATH \
```
cardigliano: Please also change the macro name to match the key name, this avoid confusion

include/ntop_defines.h
```diff
@@ -674,6 +674,8 @@
   NTOPNG_PREFS_PREFIX ".starttls"
 #define CONST_PREFS_ENABLE_DUMP_PCAP_TO_CLICKHOUSE \
   NTOPNG_PREFS_PREFIX ".dump_pcap_to_clickhouse"
+#define CONST_PREFS_ENABLE_ARCHIVE_BEFORE_TTL_DELETE \
+  NTOPNG_PREFS_PREFIX ".archive_before_ttl_delete"
```
cardigliano, line 678: Please use similar keys to clearly identify related settings (something like clickhouse_flows_archive_enabled, which is related to clickhouse_flows_archive_path)

DGabri: Fixed all, I removed the pref to specify the archive path on the GUI.
From the clickhouse daily script I expect to have the export path in this redis key: `ntopng.prefs.clickhouse_flows_archive_path` or we should rework it

PR #10051, opened 2026-02-14
Fix InfluxDB top timeseries crash when snmp_utils is unavailable

bakhtiyarneyman:

- [x] I have read the contributing guidelines https://github.com/ntop/ntopng/blob/dev/CONTRIBUTING.md

N/A

Describe changes:

Guard SNMP device lookup behind options.tags.device check in timeseries_top, matching the RRD driver's existing behavior. Previously, snmp_utils was required unconditionally for all top queries, causing a fatal Lua error for non-device schemas like iface:ndpi where no device tag is present.

### DIFF
--- a/scripts/lua/modules/timeseries/drivers/influxdb.lua
+++ b/scripts/lua/modules/timeseries/drivers/influxdb.lua
@@ -1617,7 +1617,7 @@ function driver:timeseries_top(options, top_tags)
             end
 
             local ext_label = nil
-            if options.tags.device then
+            if (ntop.isPro() and options.tags.device) then
                 local snmp_utils = require "snmp_utils"
                 local snmp_cached_dev = require "snmp_cached_dev"
                 local cached_device = snmp_cached_dev:create(options.tags.device)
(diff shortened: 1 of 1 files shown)


scripts/lua/modules/timeseries/drivers/influxdb.lua
```diff
@@ -1616,17 +1616,19 @@ function driver:timeseries_top(options, top_tags)
...
-            local cached_device = snmp_cached_dev:create(options.tags.device)
-            -- In case of flow exporters the data is port, in case of snmp it's if_index
-            local ifindex = query_tag.if_index or query_tag.port
             local ext_label = nil
-            if cached_device then
-                ext_label = snmp_utils.get_snmp_interface_label(cached_device["interfaces"][ifindex])
-            end
-            if isEmptyString(ext_label) then
-                ext_label = ifindex
+            if options.tags.device then
```
lucaderi: What about changing it to 
if (ntop.isPro() and options.tags.device) then
...

?

lucaderi (changes requested):

scripts/lua/modules/timeseries/drivers/influxdb.lua
bakhtiyarneyman: Done. FYI the check in the RRD module is for Enterprise edition, not Pro. It's out of scope for this PR, but that check could probably also be changed to `isPro`.

lucaderi: Looks good, thanks. Read isPro() as non community.

PR #10454, opened 2026-06-11
OSS-Fuzz: Add new fuzzer targets zmq processing flow

arthurscchan:
This PR creates a new OSS-Fuzz fuzzer for the ntopng project targets zmq processing flow, together with some initial corpora for the fuzzing, aiming to expand the fuzzing coverage for more ntopng logic. The new fuzzer mirror similar approach for existing OSS-Fuzz fuzzers.


fuzz/fuzz_zmq_flow.cpp
```diff
@@ -0,0 +1,242 @@
...
+    }
+  }
+
+  json_object_put(o);
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *buf, size_t len) {
+  if (len < 2) return -1;
+
+  // First byte selects the parser entry; the rest is the JSON payload
```
IvanNardi, line 199: Is the provided corpus compatible with this logic?

fuzz/fuzz_zmq_flow.options
```diff
@@ -0,0 +1,3 @@
+[libfuzzer]
+detect_leaks=0
```
IvanNardi: You likely copied that from the other fuzzer... Is that really necessary here? Did you run some tests with leak detector enabled?

fuzz/fuzz_zmq_flow.cpp
arthurscchan: Yes. All the corpora I attached has a selector byte at the beginning, following by a valid JSON payload.

fuzz/fuzz_zmq_flow.options
arthurscchan: Thanks for the comment. I did try enabling LeakSanitizer, and this fuzzer doesn't produce any leak reports in a short local run. I kept the setting only to stay consistent with the other fuzzers in the project: since a LeakSanitizer failure is reported as an OSS-Fuzz issue, I disabled it to match the existing fuzzers, in case there are unrelated leaks elsewhere in the core ntopng logic. That said, I'm happy to remove it if you'd prefer OSS-Fuzz to report memory leaks for this target as well.

IvanNardi: Unless there are "widespread" issues, I think that having it enabled allow us to swiftly detect and fix any leaks. We can always disable it later, if really needed.
My 2 cents

arthurscchan: True. I have fixed the option file.

arthurscchan:
I do trace a bit that the CIFuzz failing is from a possible real overflow problem in the code. Do you want it to be solved before this fuzzer is live? Or it could still be merged in and later allow the OSS-Fuzz to report this bug.

IvanNardi:
> Or it could still be merged in and later allow the OSS-Fuzz to report this bug

Let merge this PR as-is to add the new fuzzer and to start the real fuzzing: we can fix any issues found by it later. My 2 cents

arthurscchan:
Agree.

IvanNardi (approved):

arthurscchan:
I am happy to have this new fuzzer merged in.

PR #10555, opened 2026-07-07
guard snprintf offset overshoot in snmp handle_async_response

uwezkhan:

Describe changes:

handle_async_response builds the dotted OID string (rsp_oid[256]) and the OBJECT IDENTIFIER value (response[128]) by summing snprintf() return values into an offset that also feeds the next call's write pointer and remaining size. snprintf returns the length it would have written, so once a buffer fills the offset walks past its end and the following iteration writes at &buf[offset] while sizeof(buf) - offset wraps to a huge size_t. A response PDU whose OID (name_loc[]/name_length) or OBJECT IDENTIFIER value (val.objid[]/val_len) is long enough overflows the stack buffer, and both fields come straight from the agent through the asynch_response callback.

Before, the loop advanced the offset on any rc > 0. After, it advances only when the write fully fit (rc < remaining), the same test the hex-dump branch a few lines below already uses. Behavior is unchanged for OIDs that fit; an over-length OID now stops at the buffer edge instead of writing past it. The tradeoff is that a pathologically long OID is printed truncated rather than in full, which matches how the sibling string loop already behaves.

### DIFF
--- a/src/SNMP.cpp
+++ b/src/SNMP.cpp
@@ -108,7 +108,7 @@ void SNMP::handle_async_response(struct snmp_pdu* pdu, const char* agent_ip) {
         int rc = snprintf(&rsp_oid[offset], sizeof(rsp_oid) - offset, "%s%d",
                           (offset > 0) ? "." : "", (int)vp->name_loc[i]);
 
-        if (rc > 0)
+        if ((rc > 0) && (rc < (int)(sizeof(rsp_oid) - offset)))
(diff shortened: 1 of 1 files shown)


lucaderi:
Sorry we cannot accept your PR unless you accept the ntop Contributor License Agreement. Please accept it first.

uwezkhan:
I have accepted the ntop Contributor License Agreement.

PR #10607, opened 2026-07-17
clamp zmq_recv length in BGPPrefixListener::poll

uwezkhan:
BGPPrefixListener::poll copies BGP prefix-change publications into a 1024-byte stack buffer via zmq_recv(sock, buffer, sizeof(buffer)-1, 0). zmq_recv returns the full message size rather than the number of bytes that fit, so a publication over 1023 bytes yields a bytes_received past the buffer end, and buffer[bytes_received] = '\0' then writes off the stack.

Clamp bytes_received to sizeof(buffer)-1 before the terminator, matching the guard collect_flows already uses on its own zmq_recv. Before the change an oversized message terminates out of bounds; after it, the value is capped to what actually landed in the buffer and the trailing bytes are dropped, the same truncation the compressed-flow path accepts.

### DIFF
--- a/src/BGPPrefixListener.cpp
+++ b/src/BGPPrefixListener.cpp
@@ -110,6 +110,12 @@ void BGPPrefixListener::poll() {
       if (bytes_received < 0)
 	break;
 
+      /* zmq_recv() returns the full message size even when the message did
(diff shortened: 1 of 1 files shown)


lucaderi:
Thank you

PR #10728, opened 2026-08-09
Improve Simplified Chinese localization

Opnwall:
## What changed

- synchronize the Simplified Chinese locale with every key in the current English locale
- preserve and polish the existing Chinese translations
- add the missing SNMP, collector, and system-page translations
- correct malformed locale table nesting and align key ordering with `en.lua`

## Why

The bundled Simplified Chinese locale was substantially behind `en.lua`, leaving many ntopng 6.6 interfaces partially untranslated. A malformed collector table also placed several existing translations under incorrect key paths.

## User impact

Simplified Chinese users receive complete and more natural UI text across the current ntopng interface, while interpolation placeholders and locale lookup paths remain compatible with the English source locale.

## Validation

- compared complete leaf-key occurrence sets: 9,734 English and 9,734 Chinese, with no missing or extra keys
- confirmed shared key ordering is identical to `en.lua`
- checked interpolation placeholders for newly added entries
- parsed both locale files successfully as Lua 5.3
- ran `git diff --check`


DGabri:
@Opnwall thank you for the big contribution. Would you like a demo license for 30 days of our new Ai features and we can get your feedback? If so, please write to me at "gabriele at ntop.org"

PR #10813, opened 2026-08-23
fuzz: add SNMP response and syslog parser fuzz targets, wire flow-checks into `fuzz_dissect_packet`

IvanNardi:
- `fuzz_snmp_response`: fuzzes `SNMP::handle_async_response()` (`src/SNMP.cpp`) via net-snmp's real `snmp_parse()`, gated entirely behind `HAVE_LIBSNMP` so it only does real work on standard builds; requires `SNMP::vm` to be initialized in the constructor since the harness calls `handle_async_response()` directly without going through the normal request/response flow.
- `fuzz_syslog_parse`: fuzzes `SyslogParserInterface::parseLog()`.
- `fuzz_dissect_packet`: after the packet-dissection loop, replicate the post-pcap-read sequence PcapInterface.cpp runs in real pcap-replay mode (two full `purgeIdle()` scans + a `flow_end_housekeeping() walk`) so protocol_detected and flow_end flow-checks are deterministically exercised on every run instead of only by accident when the mutator happens to produce idle-triggering packet timestamps.


IvanNardi:
Still work-on-progress

PR #10836, opened 2026-08-26
Fix empty blacklist name in blacklisted-flow alert formatting

### LINKED ISSUE #10835: Blacklist name is always empty in blacklisted-flow alerts: `format()` reads a key that is nested one level down
## Environment

- OS: FreeBSD 15 (OPNsense 26.x), package `ntopng-6.6.d20260324_2,1`
- ntopng version: **6.6.260811, Community build**
- Interface: single packet interface, nDPI active
- Reproduced with the built-in **IPsum Threat Intelligence Feed** category list

## What happens

Every `Blacklisted Server Contact` / `Blacklisted Client Contact` alert renders the
blacklist name as an empty string, in the alert list and in every notification endpoint:

```
[Flow] [Blacklisted Server Contact] [<client>:52206 -> <server-ip>:443]
Blacklisted Server [ Blacklist: "" ]
```

The name is **not** missing from the data — it is written correctly to the alert store and
only lost at format time. The `json` column of the corresponding `flow_alerts` row:

```json
{
  "flow_risk_info": { "55": "TCP probing attempt", "50": "Connection refused (client)" },
  "blacklist": "IPsum Threat Intelligence Feed",
  "alerts": {
    "13": {
      "alert_generation": { "script_key": "blacklisted_server_contact", "subdir": "flow" },
      "cli_blacklisted": false,
      "srv_blacklisted": true,
      "cat_blacklisted": true,
      "custom_cat_file": "IPsum Threat Intelligence Feed"
    }
  }
}
```

## Root cause

`alert_utils.formatAlertMessage()` passes the **top level** of the decoded alert JSON as
`alert_type_params`:

- `alert_utils.lua` — `getAlertInfo()` returns `json.decode(alert["json"])`, the top-level table
- `alert_utils.lua` — `formatAlertMessage()` calls `description(ifid, alert, alert_json, ...)`

At the top level the list name is stored as **`blacklist`**. The key `custom_cat_file` exists
only nested under `alerts["<alert_key>"]`. But `format()` reads `custom_cat_file` from the top
level, so it is always `nil` and the template renders `""`:

- `alert_definitions/flow/alert_blacklisted_server_contact.lua` — `format()` and `add_extra_info()`
- `alert_definitions/flow/alert_blacklisted_client_contact.lua` — `format()`
- `historical_flow_utils.lua` — the `Malware` l7cat label in the historical flow view

## The codebase already disagrees with itself

In `alert_blacklisted_client_contact.lua`, the two functions read **different keys for the same
value**. `add_extra_info()` is correct; `format()` is not:

```lua
-- line 41, add_extra_info() -- correct
if alert_json and alert_json.blacklist and not isEmptyString(alert_json.blacklist) then

-- line 56, format() -- always nil
if not isEmptyString(alert_type_params["custom_cat_file"]) then
```

## Impact

The blacklist name is the single most useful field for triaging one of these alerts — it is the
difference between "a high-confidence malware feed listed this" and "one aggregator listed it
once". Without it, every blacklist alert looks identical and equally severe, and the only way to
recover the name is to query `alert_store_v11.db` by hand.

It also affects the **historical flows web UI**, where the ` @ <list name>` suffix on a Malware
category label never appears.

## Affected branches

Present in **`6.6-stable`** and **`dev`**, both checked on 26 August 2026.

## Proposed fix

Read `blacklist` first and keep `custom_cat_file` as a fallback, so nothing regresses if the key
is populated at the top level elsewhere. Patch attached / PR to follow.

jackpenny101:
Fixes #10835.

`format()` is passed the **top level** of the decoded alert JSON (`alert_utils.getAlertInfo()`
→ `formatAlertMessage()` → `description(ifid, alert, alert_json, ...)`), where the list name is
stored as `blacklist`. It read `custom_cat_file`, which exists only nested under
`alerts["<alert_key>"]`, so the value was always `nil` and every blacklisted-flow alert rendered
as:

```
Blacklisted Server [ Blacklist: "" ]
```

The name was never missing from the data — only from the formatting. Real `flow_alerts.json`
row from the alert store:

```json
{
  "blacklist": "IPsum Threat Intelligence Feed",
  "alerts": { "13": { "custom_cat_file": "IPsum Threat Intelligence Feed", "srv_blacklisted": true } }
}
```

The tree already disagreed with itself: in `alert_blacklisted_client_contact.lua`,
`add_extra_info()` reads `blacklist` while `format()` in the same file read `custom_cat_file`.

## Changes

- `alert_blacklisted_server_contact.lua` — `format()` and `add_extra_info()`
- `alert_blacklisted_client_contact.lua` — `format()`
- `historical_flow_utils.lua` — the same wrong key suppressed the ` @ <list name>` suffix on
  `Malware` category labels in the historical flow view

Each site reads `blacklist` first and keeps `custom_cat_file` as a fallback, so nothing
regresses if that key is populated at the top level elsewhere. `isEmptyString()` is nil-safe
(`ntop_utils.lua`), which the existing call sites already rely on.

Found on 6.6.260811 (Community, FreeBSD/OPNsense); the same code is present on `dev` and
`6.6-stable`.

### DIFF
--- a/scripts/lua/modules/alert_definitions/flow/alert_blacklisted_client_contact.lua
+++ b/scripts/lua/modules/alert_definitions/flow/alert_blacklisted_client_contact.lua
@@ -53,7 +53,9 @@ end
 -- @return A human-readable string
 function alert_blacklisted_client_contact.format(ifid, alert, alert_type_params)
     local blacklist = ""
-    if not isEmptyString(alert_type_params["custom_cat_file"]) then
+    if not isEmptyString(alert_type_params["blacklist"]) then
+        blacklist = alert_type_params["blacklist"]
+    elseif not isEmptyString(alert_type_params["custom_cat_file"]) then
         blacklist = alert_type_params["custom_cat_file"]
     end
     -- The client is blacklisted, so simply use the Client as who description    
--- a/scripts/lua/modules/alert_definitions/flow/alert_blacklisted_server_contact.lua
+++ b/scripts/lua/modules/alert_definitions/flow/alert_blacklisted_server_contact.lua
@@ -38,8 +38,9 @@ end
 -- ##############################################
 
 function alert_blacklisted_server_contact:add_extra_info(alert_json)
-   if alert_json and alert_json.custom_cat_file and not isEmptyString(alert_json.custom_cat_file) then
-        return " [ " .. i18n("flow_details.blacklist", { blacklist = alert_json.custom_cat_file or "" }) .. " ] "
+   local blacklist = alert_json and (alert_json.blacklist or alert_json.custom_cat_file)
+   if not isEmptyString(blacklist) then
+        return " [ " .. i18n("flow_details.blacklist", { blacklist = blacklist }) .. " ] "
     end
     return ""
 end
@@ -53,7 +54,9 @@ end
 -- @return A human-readable string
 function alert_blacklisted_server_contact.format(ifid, alert, alert_type_params)
     local blacklist = ""
-    if not isEmptyString(alert_type_params["custom_cat_file"]) then
+    if not isEmptyString(alert_type_params["blacklist"]) then
+        blacklist = alert_type_params["blacklist"]
+    elseif not isEmptyString(alert_type_params["custom_cat_file"]) then
         blacklist = alert_type_params["custom_cat_file"]
     end
     -- The client is blacklisted, so simply use the Client as who description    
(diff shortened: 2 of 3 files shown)


IvanNardi:
@jackpenny101 could you sign the CLA, please and tick the box(es), please?

- [] I have signed the ntop Contributor License Agreement at https://github.com/ntop/legal/blob/main/individual-contributor-licence-agreement.md
- [] I have read the contributing guide lines https://github.com/ntop/ntopng/blob/dev/CONTRIBUTING.md
- [] I have updated the documentation (in doc/src/) to reflect the changes made (if applicable)

jackpenny101:
Withdrawing this — I'm not in a position to sign the CLA, so I'd rather not leave it sitting in your queue.

The underlying issue is real and reproducible if anyone wants to pick it up: `format()` reads `alert_type_params["custom_cat_file"]`, but `formatAlertMessage()` passes the top level of the alert JSON, where the key is `blacklist`. Details are in #10835. Apologies for the noise.

PR #10860, opened 2026-08-28
bound snmp asn1 value length to remaining bytes

uwezkhan:

Describe changes:

The built-in SNMP client parses agent responses with a small ASN.1 reader. asn1_parse_string_type, asn1_parse_integer_type and asn1_parse_oid take the BER content length from next_len() and then read that many bytes off the datagram, but unlike asn1_parse_structure they never check the length against parser->state->remaining. A response whose OCTET STRING, INTEGER or OID declares more bytes than the packet carries makes the read loop run past the buffer. snmp_parse_message reaches asn1_parse_string for the community field before any varbind, so an 8 byte datagram 30 06 02 01 00 04 7F 41 (community length 0x7F, one byte present) is enough. This is the default build path, since HAVE_LIBSNMP is off unless net-snmp is installed.

Before, the three value parsers trusted the wire length; after, each rejects a length that would extend past the remaining bytes, reusing the same containment test asn1_parse_structure already performs. The check sits in the leaf parsers because next_len() cannot see which read follows and the callers above do not know the field width. Valid responses produce identical output; the tradeoff is that a forged or truncated length now aborts that item rather than returning bytes read from adjacent memory.

### DIFF
--- a/third-party/snmp/asn1.c
+++ b/third-party/snmp/asn1.c
@@ -510,7 +510,18 @@ int asn1_parse_integer_type(ASN1Parser *parser, int *type, int64_t *dest)
     
   size = next_len(parser);
   payload = next_payload(parser);
-    
+
+  /* A declared length must stay within the bytes still present in the current
+     state, otherwise the read loop walks off the datagram. asn1_parse_structure
+     already performs this containment check on its payload. */
+  if ((size < 0) ||
+      ((char*)payload + size >
+       (char*)parser->state->buffer + parser->state->remaining))
+    {
+      if (dest) *dest = 0;
+      return 0;
+    }
+
   for (i = 0; i < size; i++)
     {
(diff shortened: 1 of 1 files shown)


IvanNardi:

- [ ] I have signed the ntop Contributor License Agreement at https://github.com/ntop/legal/blob/main/individual-contributor-licence-agreement.md
- [ ] I have read the contributing guide lines https://github.com/ntop/ntopng/blob/dev/CONTRIBUTING.md

uwezkhan:
Ticked the CLA and contributing guidelines boxes in the PR body. The agreement was already signed on my end, same account as #10681, and the diff is unchanged.

IvanNardi (approved):
LGTM

IvanNardi:
Thank you

PR #10865, opened 2026-08-30
Fix radius updateLoginInfo early return

StefanJum:

Describe changes:

The Radius::updateLoginInfo checks if any of the auth or acct radius servers are set to empty strings, and if to, returns false. This leads to the "use_chap" variable never being properly set. This can lead to the wrong auth protocol being set if the acct server is not set.

The fix is just to move the auth protocol check before the check for both auth and acct servers.


IvanNardi:
@StefanJum, could you sign the CLA, please?

StefanJum:
Done @IvanNardi , sorry, I forgot to submit earlier.

IvanNardi (approved):
LGTM

IvanNardi:
Thank you

PR #10883, opened 2026-09-01
OSS-Fuzz: Add fuzzer and cropus targets mdns

arthurscchan:
This PR creates a new fuzzer and corpus targets mdns dissect process.


IvanNardi:
@arthurscchan, could you try to provide a larger corpus, with different types of payload, please?

arthurscchan:
Ok. I will handle that, sorry been kept off for a few days. Will do it on next week.

PR #10913, opened 2026-09-09
bound detunneled eth_offset to caplen in dissectPacket

uwezkhan:

N/A

Describe changes:

With tunnel decoding enabled (-T), the VXLAN, EOIP, GRE Transparent Ethernet Bridging/ERSPAN, and TZSP decoders set `eth_offset` and `goto datalink_check` to re-parse the inner Ethernet frame. Those decoders can leave `eth_offset` equal to or past `h->caplen`: VXLAN and EOIP add their header lengths with no bounds check, and the GRE TEB/ERSPAN path only checks `h->caplen >= offset`. The `DLT_EN10MB` branch then validates the total `h->caplen` against `sizeof(ndpi_ethhdr)` rather than against `eth_offset`, and reads `ethernet->h_proto` at `packet[eth_offset + 12]`. A short tunneled UDP datagram that ends right after the outer UDP header reads past the captured packet buffer. The Linux cooked (113/276) and DLT_RAW branches carry the same `eth_offset`-relative reads guarded only by total `caplen`.

Before, only the `DLT_NULL` branch guarded its read with `eth_offset + size <= h->caplen`; after, the four other branches use the same form. Keeping the bound in the shared datalink parser covers every tunnel decoder that re-enters it, present and future, instead of repeating the check at each `goto`. The only behavioral trade-off is for `eth_offset == 0` (non-tunneled traffic), where the new condition is identical to the old one, so normal captures are unaffected.

### DIFF
--- a/src/NetworkInterface.cpp
+++ b/src/NetworkInterface.cpp
@@ -3054,7 +3054,7 @@ datalink_check:
     if (sender_mac) memcpy(&dummy_ethernet.h_source, sender_mac, 6);
     ip_offset = 4 + eth_offset;
   } else if (datalink_type == DLT_EN10MB) {
-    if (h->caplen < sizeof(ndpi_ethhdr)) {
+    if ((eth_offset + sizeof(ndpi_ethhdr)) > h->caplen) {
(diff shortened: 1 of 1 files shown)


IvanNardi (approved):
LGTM

PR #10937, opened 2026-09-15
Reject markup characters in host names learnt from the network

kah-ja:
### What

`Host::isValidHostName` (`src/Host.cpp:1797`) rejects `:`, `*`, space, `@`, `,`, `(` and `)`, but accepts `<`, `>`, `"` and `'`. `Host::setResolvedName` (`src/Host.cpp:1826`) runs no check at all, so a reverse DNS answer becomes the stored name of a host verbatim (`src/Flow.cpp:1116,1152`, `src/NetworkInterface.cpp:5202`).

### Why

Every other name setter in that file already validates: `offlineSetMDNSName:1744`, `offlineSetDHCPName:1751`, `offlineSetNetbiosName:1765`, `offlineSetTLSName:1773`, `offlineSetHTTPName:1789`, `setServerName:1818`. The resolved name is the one path that skips it. The stored names are handed to the GUI as they were seen on the wire, for example by `scripts/lua/rest/v2/get/host/active_list.lua:171-174` as `hostname.name`, and `ntop.httpPurifyParam` never applies to them because they do not arrive as HTTP parameters. None of the four characters is valid in a host name, so no legitimate name is lost.

### How

Two changes in the same family:

- `isValidHostName` also rejects `<`, `>`, `"` and `'`.
- `Host::setResolvedName` runs `isValidHostName` first, like its siblings, and the `LocalHost` override (`src/LocalHost.cpp:811`) does the same before it records the name through `addDataToAssets`, so an invalid name is not kept as asset data either.

The GUI side is deliberately untouched: the templates now receive names that cannot carry markup, and the Vue sources are no longer part of this repository.

### Testing

- The predicate was exercised standalone (same code, compiled with `g++ -std=c++17 -Wall -Wextra`). Before the change ``, `<svg/onload=x>`, `x"onmouseover="x` and `'-x-'` were all accepted; after it they are rejected. `printer.lan`, `NAS-01.fritz.box`, `host-with-dash_1.example.com` and `xn--bcher-kva.example` stay accepted, `1.2.3.4` stays rejected as before.
- ntopng itself was not built or run for this change; the repository has no unit tests for the host name path (only the `tests/e2e` submodule).

### References

GHSA-89wv-w893-9fvq (reported privately, unpublished at time of writing)

Found during a [penetration test](https://www.turingpoint.de/en/security-assessments/pentests/) by turingpoint.


lucaderi:
Thanks for your contribution but I have decided to implement several validators, one per protocol, rather than using a generic one as you did. Hence due to https://github.com/ntop/ntopng/commit/3ef23ce6ce88e6712b87b93c9a06435df6478728 I close you PR and invite to review the code

